PR #16, opened 2019-01-30
[WIP] add routing

damienomurchu:
**Not completed or ready to merge**

## Motivation
Add dedicated routing file for endpoints https://issues.jboss.org/browse/AEROGEAR-8456

## What
Routing added & wired up

## Why
Support addition of endpoints

## How
<TODO>

## Verification Steps
<TODO>
## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


damienomurchu:
@austincunningham yeah, I'm working off unmerged changes from other PR's, so this branch will often be out of sync :)

pkg/config/config.go
```diff
@@ -28,6 +30,8 @@ func Get() Config {
 			AllowOrigins:     getEnvSlice("ACCESS_CONTROL_ALLOW_ORIGIN", []string{"*"}, ","),
 			AllowCredentials: getEnvBool("ACCESS_CONTROL_ALLOW_CREDENTIALS", false),
 		},
+		StaticFilesDir:  getEnv("STATIC_FILES_DIR", ""),
+		ApiRoutePrefix: "/api", //should start with a "/",
```
damienomurchu: I'm wondering should this perhaps be `api/v1` to version the api?

craicoverflow: Yeah it definitely should. It could even be separated into a separate config variable for maximum flexibility.

```go
...
ApiRoutePrefix: "/api", //should start with a "/",
ApiVersionPrefix: "/v1"
...
```

pkg/web/router.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (v *RequestValidator) Validate(i interface{}) error {
+	return v.validator.Struct(i)
+}
+
+func NewRouter(fileDir string, apiRoutePrefix string) *echo.Echo {
+	router := echo.New()
+
+	router.Use(middleware.Logger())
```
davidffrench: @damienomurchu The middleware is getting loaded twice, once here and in main.go. It should just occur here, using middleware.Init in my opinion.

pkg/config/config.go
```diff
@@ -28,6 +30,8 @@ func Get() Config {
 			AllowOrigins:     getEnvSlice("ACCESS_CONTROL_ALLOW_ORIGIN", []string{"*"}, ","),
 			AllowCredentials: getEnvBool("ACCESS_CONTROL_ALLOW_CREDENTIALS", false),
 		},
+		StaticFilesDir: getEnv("STATIC_FILES_DIR", ""),
+		ApiRoutePrefix: "/api/v1", //should start with a "/",
```
davidffrench: should split the version into a separate config property that can be read from an environment variable and defaults to v1.

cmd/mobile-security-service/main.go
```diff
@@ -57,12 +62,15 @@ func initLogger(level, format string) {
...
+func initHandlers(e *echo.Echo, c config.Config) {
 	// App handler setup
 	appsPostgreSQLRepository := apps.NewPostgreSQLRepository()
 	appsService := apps.NewService(appsPostgreSQLRepository)
 	appsHandler := apps.NewHTTPHandler(e, appsService)
 
-	// Define /app routes
-	e.GET("/apps", appsHandler.GetApps)
+	// Setup /app routes
+	apiRoutePrefix := c.ApiRoutePrefix
```
davidffrench: the apiRoutePrefix and apiGroup should move above the app handler setup as these will be used for all routes.

cmd/mobile-security-service/main.go
```diff
@@ -23,12 +24,16 @@ func init() {
...
 	config := config.Get()
-	e := echo.New()
+	staticFilesDir := config.StaticFilesDir
+	apiRoutePrefix := config.ApiRoutePrefix
+
+	//e := echo.New()
+	e := web.NewRouter(staticFilesDir, apiRoutePrefix)
 
 	// Load middleware
 	middleware.Init(e, config)
```
davidffrench: Move this, based on comment in router.go

pkg/config/config.go
damienomurchu: Agree - done

pkg/config/config.go
damienomurchu: Split into 2 config properties now - just confirming, do we want the api version to be able to be set by env var?

pkg/config/config.go
```diff
@@ -28,6 +31,9 @@ func Get() Config {
 			AllowOrigins:     getEnvSlice("ACCESS_CONTROL_ALLOW_ORIGIN", []string{"*"}, ","),
 			AllowCredentials: getEnvBool("ACCESS_CONTROL_ALLOW_CREDENTIALS", false),
 		},
+		StaticFilesDir: getEnv("STATIC_FILES_DIR", ""),
+		ApiRoutePrefix: "/api", //should start with a "/",
+		ApiVersionPrefix: "/v1",
```
davidffrench: Now that I think about it, this is not going to be enough for versioning in my opinion. It is not enough to be hardcoded and also not enough to have as an environment variable. That will just change the string in the endpoint but not actually achieve any versioning at all. There is also a consideration for the OAuth proxy and how it can handle a potentially dynamic URL.

I would remove this entirely for now and we can tackle it later.

pkg/web/router.go
```diff
@@ -0,0 +1,41 @@
...
+			// We don't want to return the SPA if any api/* is called, it should act like a normal API.
+			return strings.HasPrefix(context.Request().URL.Path, apiRoutePrefix)
+		},
+		Browse: false,
+	}))
+	router.Validator = &RequestValidator{validator: validator.New()}
+	return router
+}
+
+func SetupAppsRoute(r *echo.Group, appsHandler *apps.HTTPHandler) {
```
craicoverflow: I think this function name could be changed to `SetAppRoutes`. It is slightly shorter and also we are defining **multiple routes** so it should be plural.

pkg/web/router.go
```diff
@@ -0,0 +1,41 @@
+package web
```
craicoverflow: I think it would be best to give the router its own package. So it could be called like:

```go
r := router.New()
```

It could be added under **pkg/web/router**.

cmd/mobile-security-service/main.go
damienomurchu: working on this now

pkg/config/config.go
damienomurchu: resolved by your below reply

cmd/mobile-security-service/main.go
craicoverflow: @davidffrench it will work regardless.

craicoverflow:
You should include the ticket in the commit message

pkg/config/config.go
damienomurchu: Removed entirely now :+1:

damienomurchu:
@craicoverflow will add ticket ref when I squash :+1:

PR #21, opened 2019-02-01
fix: Change location of api spec and remove fields from App

davidffrench:
## Motivation
No JIRA

## What
Change the location of api spec and removing disabled and disabledMessage fields from App level

## Why
Adhering to Go folder structure by moving swagger spec file to api folder. 
Removing disabled and disabledMessage fields due to change in UI.

## Verification Steps
Not possible to verify the changes until they are in GitHub :-(

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Now the application will use the values defined in `.env`.
 A [Swagger](https://swagger.io/) UI can be used for testing the mobile-security-service service.
 
 ```bash
-docker run -p 8080:8080 -e API_URL=https://raw.githubusercontent.com/aerogear/mobile-security-service/master/apispec.yaml swaggerapi/swagger-ui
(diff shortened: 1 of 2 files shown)


davidffrench:
> Commit message isn't adhering to our new standard:
> 
> `fix(scope): Change location of api spec and remove fields from App (AEROGEAR-XXXX)`

I have found the first problem with the commit message format. If there is no JIRA associated. @laurafitzgerald What is your opinion on always having a JIRA for every change?

craicoverflow:
> > Commit message isn't adhering to our new standard:
> > `fix(scope): Change location of api spec and remove fields from App (AEROGEAR-XXXX)`
> 
> I have found the first problem with the commit message format. If there is no JIRA associated. @laurafitzgerald What is your opinion on always having a JIRA for every change?

This is also relevant to @austincunningham as he is working on the CircleCI regex rule for this.

austincunningham:
(AEROGEAR-XXXX) will satisfy the rule at present

davidffrench:
Modified now. I change it to (none) first but I didn't like it that way.

PR #23, opened 2019-02-04
feat(commit-message-check): adding bash script for checking commit messages to circle ci (AEROGEAR-8465)

austincunningham:
## Motivation
adding bash script for checking commit messages to circle ci 
https://issues.jboss.org/browse/AEROGEAR-8465

## What
adding bash script for checking commit messages to circle ci 

## Why
to meet with the guide lines outline here
https://github.com/fheng/help/blob/master/development_process.md#commit-messages

## How
 adding bash script for checking commit messages to circle ci 
## Verification Steps
If this PR passes the circle ci build the you can consider it working

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


scripts/commit-filter-check.sh
```diff
@@ -0,0 +1,32 @@
...
+git log $currentbranch --pretty=format:"%H" --not master > shafile.txt
+
+# loops through the file an gets the message
+for i in `cat ./shafile.txt`;
+do 
+  # gets the git commit message based on the sha
+  gitmessage=`git log --format=%B -n 1 "$i"`
+
+  # fix(commit-filter-check): add commit messages (AEROGEAR-038928990423)
+   messagecheck=`echo $gitmessage | grep -w "feat\|fix\|docs" | grep "(AEROGEAR-" | grep "): "`
```
davidffrench: This does not seem to be working for me. I am testing it with `feat: sdasdsadsaas (AEROGEAR-asdsada)`

scripts/commit-filter-check.sh
```diff
@@ -0,0 +1,32 @@
...
+
+  # check to see if the messagecheck var is empty
+  if [ -z "$messagecheck" ]
+  then 
+        echo "$messagecheck"
+        echo "$gitmessage"
+        echo "failed commit message check"
```
davidffrench: Would be good to output why it failed, I know this adds extra complexity. Or at least list what the structure requirement is here.

scripts/commit-filter-check.sh
austincunningham: I would expect that to fail doesn't match the format
` feat(some txt): some txt (AEROGEAR-sometxt)`
Have a test repo on my own for different commits https://github.com/austincunningham/-git-commit-messages-filtering.git, checkout the branchname, didn't want to clutter production with test commit messages.

scripts/commit-filter-check.sh
austincunningham: will see what I can do

scripts/commit-filter-check.sh
davidffrench: The scope section is optional, i would have expected the above to pass.

scripts/commit-filter-check.sh
```diff
@@ -13,14 +13,37 @@ do
   gitmessage=`git log --format=%B -n 1 "$i"`
 
   # fix(commit-filter-check): add commit messages (AEROGEAR-038928990423)
-   messagecheck=`echo $gitmessage | grep -w "feat\|fix\|docs" | grep "(AEROGEAR-" | grep "): "`
+  messagecheck=`echo $gitmessage | grep -w "feat\|fix\|docs"`
```
davidffrench: Looking good. It happened to output all checks even though it only failed one. Is this happening for you? Again, I was using `feat: sdasdsadsaas (AEROGEAR-asdsada)`

scripts/commit-filter-check.sh
austincunningham: As its optional then no point checking will remove the check

scripts/commit-filter-check.sh
austincunningham: was seeing that issue when using ` instead of " for the string I was passing in. Not it seeing now

davidffrench: seems to be working fine for me now, weird.

PR #28, opened 2019-02-06
Aerogear 8503

davidffrench:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8503

## What
- Changing Swagger docs after ERD and router changes
- Fixing other minor errors in the code

## Why
Swagger docs are out of date

## Verification Steps
1. Install [Chrome extension](https://chrome.google.com/webstore/detail/swagger-ui-console/ljlmonadebogfjabhkppkoohjkjclfai?hl=en)
2. Paste [https://gist.githubusercontent.com/davidffrench/61a85d43bf2d526380bbc842960312d5/raw/b8e478b8e0efe547f48b1fbfe735187dbc92c71a/swagger.yml](https://gist.githubusercontent.com/davidffrench/61a85d43bf2d526380bbc842960312d5/raw/b8e478b8e0efe547f48b1fbfe735187dbc92c71a/swagger.yml) and press **Explore**

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


davidffrench:
@austincunningham They do, if you look at the bottom of the screenshot, you will see `BASE URL: /api`

austincunningham:
@davidffrench yea seen that but its a pet hate of mine with api docs that you never get the full route in the same place

api/swagger.yaml
```diff
@@ -151,9 +146,8 @@ definitions:
   AppInit:
     type: object
     properties:
-      id:
-        type: integer
-        format: int64
+      appId:
+        type: string
```
damienomurchu: Can I ask why the switch from integer to string & how this is being generated?

craicoverflow: > @davidffrench yea seen that but its a pet hate of mine with api docs that you never get the full route in the same place

@davidffrench @austincunningham I don't think `/api` is the base URL - it is the base path. Something like `http://my-server-addr.com` would be the base URL.

craicoverflow:
> @austincunningham They do, if you look at the bottom of the screenshot, you will see `BASE URL: /api`

@davidffrench @austincunningham I don't think `/api` is the base URL - it is the base path. Something like `http://my-server-addr.com` would be the base URL.

api/swagger.yaml
davidffrench: OF course, the appId will be in the format `A!B2C3D4E5.com.domainname.applicationname`

davidffrench:
> > @austincunningham They do, if you look at the bottom of the screenshot, you will see `BASE URL: /api`
> 
> @davidffrench @austincunningham I don't think `/api` is the base URL - it is the base path. Something like `http://my-server-addr.com` would be the base URL.

You are right, it was just the easiest way of putting it in the comment. In the changed doc, the property is `basePath`. I have no host or base url set. In the screenshot, swagger combines the host and basePath to list the `base url` at the bottom.

craicoverflow:
screenshot from 2019-02-06 10-44-19

Notice under **Response Messages** you list a 200 on **PUT** `/apps/{id}` but this isn't listed in the other endpoints (only 4xx's)

davidffrench:
> screenshot from 2019-02-06 10-44-19
> 
> Notice under **Response Messages** you list a 200 on **PUT** `/apps/{id}` but this isn't listed in the other endpoints (only 4xx's)

Swagger shows it in a strange way. The other ones do have a 200 response but since they have a response body, it shows it differently. This PUT does not have a response body.

PR #31, opened 2019-02-06
feat: adding initial react-app structure

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8377

## What
Initial React App structure

I pushed the entire setup but I've left a self-review on some of the files which I think we can exclude from the initial setup for now. See comments below

## Why
Add an short answer for: Why it was done? (E.g The feature X was deprecated.)

## How
Using create-react-app 

## Verification Steps
Follow Local Development Steps in the README file of this pr.
 
**Expected outcome** is a running react app in debug mode of vscode

## Checklist:


ui/src/serviceWorker.js
```diff
@@ -0,0 +1,135 @@
+// This optional code is used to register a service worker.
```
laurafitzgerald: Not sure we need this one for now.

ui/yarn.lock
```diff
@@ -0,0 +1,10051 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
laurafitzgerald: I also don't think there's a need to check this one as we are using npm to run the application.

ui/README.md
```diff
@@ -0,0 +1,14 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
davidffrench: I think we should remove this line.

ui/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Local Development using VSCode Chrome Debugger
+
+Run:
+
+### `npm start`
```
davidffrench: I have tested it but `npm start` shouldn't be needed to run in VS Code.

ui/src/serviceWorker.js
davidffrench: no need for offline capability for us. Would be cool though :-)

ui/yarn.lock
davidffrench: I believe we can switch to just using npm fully rather than yarn. This should be removed and a package.lock file created instead in my opinion.

wtrocki:
yarn.lock can be added to gitignore.

ui/src/serviceWorker.js
wtrocki: It is backbone anyway and gives application shel for production builds only.
Removing this may affect production builds that will look for serviceWorker in root folder.
It should be removed with manifest.

ui/README.md
craicoverflow: It's needed for now @davidffrench as the `.vscode/launch.json` does not exist yet. I think it should be kept in regardless as not everyone uses VS Code (our team do, but this is a public project so there will be others who want to run this).

ui/README.md
laurafitzgerald: Will do, just wanting to give the nod to them.

ui/README.md
laurafitzgerald: @davidffrench @craicoverflow with the launch.json in place I'm able to run in debug mode, but the application is unavailable until I run `npm start`. I could be missing a step somehow. The file is checked in now so would you mind checking that on your end, please. 😄

craicoverflow: @laurafitzgerald just finished testing that. I think that is the expected behaviour as the `launch.json` does not actually start the server, it binds to the URL which the server is running on already, so it needs to be left in. Maybe @davidffrench could confirm.

davidffrench: Ye are absolutely right. You need to run `npm start`. https://code.visualstudio.com/docs/nodejs/reactjs-tutorial#_configure-the-chrome-debugger

davidffrench:
> yarn.lock can be added to gitignore.

@wtrocki There should be no need for yarn at all in my opinion. What is your take on this?

ui/README.md
davidffrench: That would be the nice thing to do I suppose :-)

ui/README.md
craicoverflow: This is a h3 type heading, I don't think it should be. The `###` should be removed and it should be converted it to a code block like below. See [Code and Syntax Highlighting](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting)

```sh
npm start
```

ui/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Runs the app in the development mode on localhost:3000<br>
+Opem the project in VSCode
```
craicoverflow: "Ope**n**" should be ""Ope**n**"

ui/README.md
```diff
@@ -0,0 +1,14 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
+
+## Local Development using VSCode Chrome Debugger
+
+Run:
+
+### `npm start`
+
+Runs the app in the development mode on localhost:3000<br>
```
craicoverflow: I think these steps should be converted to an ordered list so it is easier to follow.

davidffrench:
> Some small changes I think need to be made before this can be approved.
> 
> Also I think if we are changing the README for the API server to AsciiDoc should this README also be the same? @damienomurchu @camilamacedo86 @austincunningham @davidffrench

Yes, we should start as we mean to go on.

craicoverflow:
Looks good @laurafitzgerald ,I've tested locally. One question - have you checked in the `package-lock.json`? I can't see it but one generated for me on `npm install`.

> > yarn.lock can be added to gitignore.
> 
> @wtrocki There should be no need for yarn at all in my opinion. What is your take on this?

I guess it should be added to gitignore regardless of if we choose to use it or not. Individual people may have a preference to use it for their own reasons, so if they choose to then it should be catered for by including it in the gitignore

craicoverflow:
The root `.gitignore` file should be updated to include the following line:

```
ui/node_modules
```

Otherwise the node_modules will be checked in.

craicoverflow:
We are now managing two `.gitignore` files on different levels ~ should be merge the contents of `ui/.gitignore` into that of the root project?

@davidffrench @austincunningham @camilamacedo86 @damienomurchu @laurafitzgerald

laurafitzgerald:
> We are now managing two `.gitignore` files on different levels ~ should be merge the contents of `ui/.gitignore` into that of the root project?
> 
> @davidffrench @austincunningham @camilamacedo86 @damienomurchu @laurafitzgerald

@craicoverflow I think we should yes. I followed the structure of the mobile-developer-console here but as you've pointed out this creates the need to manage two. I'm 👍 on putting it all in the root one.

ui/README.md
damienomurchu: +1 on removing this line, but we can do that when we refactor README I think

ui/src/App.js
```diff
@@ -0,0 +1,28 @@
...
+          </p>
+          <a
+            className="App-link"
+            href="https://reactjs.org"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            Learn React
+          </a>
+        </header>
```
damienomurchu: Can remove below & just leave div tags
```
<header className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
          <p>
            Edit <code>src/App.js</code> and save to reload.
          </p>
          <a
            className="App-link"
            href="https://reactjs.org"
            target="_blank"
            rel="noopener noreferrer"
          >
            Learn React
          </a>
        </header>
```

ui/src/App.js
```diff
@@ -0,0 +1,28 @@
+import React, { Component } from 'react';
+import logo from './logo.svg';
```
damienomurchu: Can remove this as well, as we won't be using referenced logo

ui/src/logo.svg
```diff
@@ -0,0 +1,7 @@
...
+        <circle cx="420.9" cy="296.5" r="45.7"/>
+        <path d="M520.5 78.1z"/>
+    </g>
+</svg>
```
damienomurchu: Can remove this entire image file as we won't be using it in our app

ui/src/serviceWorker.js
```diff
@@ -0,0 +1,135 @@
...
+    });
+}
+
+export function unregister() {
+  if ('serviceWorker' in navigator) {
+    navigator.serviceWorker.ready.then(registration => {
+      registration.unregister();
+    });
+  }
+}
```
damienomurchu: Is serviceworker implementation code in this file relevant to our mobile-security-service? If unsure, would remove for now

.gitignore
```diff
@@ -1,9 +1,39 @@
...
+ui/.env.test.local
+ui/.env.production.local
+
+ui/npm-debug.log*
+ui/yarn-debug.log*
+ui/yarn-error.log*
+ui/yarn.lock
+
+ui/.vscode/*
+!ui/.vscode/launch.json
```
damienomurchu: Can we use some grepping patterns above, to make the .gitignore less verbose, ie `*.log`, etc

ui/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
damienomurchu: Change the names here

ui/src/App.css
```diff
@@ -0,0 +1,32 @@
...
+}
+
+@keyframes App-logo-spin {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
```
damienomurchu: We will probably remove/ change all of this css, so we could remove this file now

damienomurchu:
as @craicoverflow mentioned, we are agreed on using npm instead of yarn, and we should check in a `package-lock.json` file (you can do this by running `npm install` locally to generate it & then commit it)

.gitignore
laurafitzgerald: @damienomurchu not sure what change you are suggesting here. The goal of line 38 and 39 is to only include launch.json from the .vscode folder.

ui/src/serviceWorker.js
laurafitzgerald: @damienomurchu will remove. There seems to be consensus on removing it.

ui/src/logo.svg
laurafitzgerald: @damienomurchu will do. I removed the reference to it i App.js as well.

ui/src/App.js
laurafitzgerald: 👍

.gitignore
damienomurchu: @laurafitzgerald actually meant the increasingly verbose `.gitignore` file, not just lines 38/ 39 (github only seems to let me comment on one line :( ). What I was trying to say was that we might be able to reduce the number of entries added to the `.gitignore` by using grepping patterns, ie like adding `*.log` to ignore all .log files rather than adding each.log file to be ignored, if that makes sense? :)

laurafitzgerald:
@damienomurchu @craicoverflow @davidffrench thanks for the reviews. I've addressed the feedback and pushed changes. I'd like to squash all the changes to one commit but want to check if you guys try to split changes. I think these all fit into the one change of adding the initial setup. Will wait for feedback before squashing.

ui/package-lock.json
```diff
@@ -0,0 +1,17601 @@
+{
+  "name": "ui",
```
damienomurchu: @laurafitzgerald just noticed this now - should we rename this from `ui` to `mobile-security-service-ui`? (maybe we don't need to, I'm just not sure)

ui/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "ui",
```
damienomurchu: @laurafitzgerald just noticed this now - should we rename this from `ui` to `mobile-security-service-ui`? (maybe we don't need to, I'm just not sure)

PR #34, opened 2019-02-08
Adding generic errors and mapping to http errors

davidffrench:
## Motivation
Our clean architecture separates business logic (filename=service) and delivery (http). We should not be using http errors in the services as multiple delivery methods (http, CLI) would be using the same service file.

## What
Added generic error model that can be used in both the service and repository files. This is then mapped to the http errors which will only be done at the http delivery layer.

## Why
Clean separation of responsibilities between our layers. 

## How
All errors in the service and repository layer will return an error using the method `return models.ErrInternalServerError`

## Verification Steps
None needed in my opinion. Tests are written to cover all cases.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/httperrors/httperrors.go
+++ b/pkg/httperrors/httperrors.go
@@ -3,6 +3,7 @@ package httperrors
 import (
 	"encoding/json"
 
+	"github.com/aerogear/mobile-security-service/pkg/models"
 	"github.com/labstack/echo"
 )
 
@@ -121,4 +122,23 @@ func HTTPError(c echo.Context, statusCode int, message string) (e error) {
 	}
 
 	return c.JSON(statusCode, resBody)
+}
+
+// GetHTTPResponseFromErr returns the mapped http error to the generic errors model.
+func GetHTTPResponseFromErr(c echo.Context, err error) (e error) {
+
+	switch err {
+	case models.ErrInternalServerError:
+		return InternalServerError(c, err.Error())
+	case models.ErrNotFound:
+		return NotFound(c, err.Error())
+	case models.ErrConflict:
+		return Conflict(c, err.Error())
+	case models.ErrBadParamInput:
+		return BadRequest(c, err.Error())
+	case models.ErrUnauthorized:
+		return Unauthorized(c, err.Error())
+	default:
+		return InternalServerError(c, "")
+	}
 }
--- a/pkg/httperrors/httperrors_test.go
+++ b/pkg/httperrors/httperrors_test.go
@@ -1,11 +1,13 @@
 package httperrors
 
 import (
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
 	"testing"
 
+	"github.com/aerogear/mobile-security-service/pkg/models"
 	"github.com/labstack/echo"
 )
 
@@ -37,6 +39,60 @@ func TestHTTPError(t *testing.T) {
 
 			if _ = HTTPError(c, tt.args.statusCode, tt.args.message); !reflect.DeepEqual(rec.Code, tt.wantCode) {
 				t.Errorf("HTTPError() error = %v, wantErr %v", rec.Code, tt.args.statusCode)
+			}
+		})
+	}
+}
+
+func TestGetHTTPResponseFromErr(t *testing.T) {
+
+	tests := []struct {
+		name     string
+		args     error
+		wantCode int
+	}{
+		{
+			name:     "GetHTTPResponseFromErr() should return a HTTP error with a 500 status code",
+			args:     models.ErrInternalServerError,
+			wantCode: http.StatusInternalServerError,
+		},
+		{
+			name:     "GetHTTPResponseFromErr() should return a HTTP error with a 404 status code",
+			args:     models.ErrNotFound,
+			wantCode: http.StatusNotFound,
+		},
+		{
+			name:     "GetHTTPResponseFromErr() should return a HTTP error with a 409 status code",
+			args:     models.ErrConflict,
+			wantCode: http.StatusConflict,
+		},
+		{
+			name:     "GetHTTPResponseFromErr() should return a HTTP error with a 400 status code",
+			args:     models.ErrBadParamInput,
+			wantCode: http.StatusBadRequest,
+		},
+		{
+			name:     "GetHTTPResponseFromErr() should return a HTTP error with a 401 status code",
+			args:     models.ErrUnauthorized,
+			wantCode: http.StatusUnauthorized,
+		},
+		{
+			name:     "GetHTTPResponseFromErr() should return a default HTTP error with a 500 status code",
+			args:     errors.New(""),
+			wantCode: http.StatusInternalServerError,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create a mock echo Context
+			e := echo.New()
+			req := httptest.NewRequest(http.MethodGet, "/", nil)
+			rec := httptest.NewRecorder()
+			c := e.NewContext(req, rec)
+
+			if _ = GetHTTPResponseFromErr(c, tt.args); !reflect.DeepEqual(rec.Code, tt.wantCode) {
+				t.Errorf("HTTPError() error = %v, wantErr %v", rec.Code, tt.wantCode)
 			}
 		})
 	}
--- a/pkg/web/apps/apps_http_handler.go
+++ b/pkg/web/apps/apps_http_handler.go
@@ -3,6 +3,9 @@ package apps
 import (
 	"net/http"
 
+	"github.com/aerogear/mobile-security-service/pkg/models"
+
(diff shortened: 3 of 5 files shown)


pkg/httperrors/httperrors.go
```diff
@@ -122,3 +123,22 @@ func HTTPError(c echo.Context, statusCode int, message string) (e error) {
...
 }
+
+// GetHTTPResponseFromErr returns the mapped http error to the generic errors model.
+func GetHTTPResponseFromErr(c echo.Context, err error) (e error) {
+
+	switch err {
+	case models.ErrInternalServerError:
+		return InternalServerError(c, err.Error())
+	case models.ErrNotFound:
+		return NotFound(c, err.Error())
```
camilamacedo86: Hi @davidffrench ,

Is the error "models.ErrNotFound" implemented to be returned when a **DATA** is NOT FOUND? If yes, the HTTP code 404 returned here according to the HTTP definitions should be used to indicates that the client (web browser) is receiving a message from the server (remote computer) that the specific resource (web page/URL) is unavailable. It is not a valid code when DATA in the database is not found/do not exist. 

Following the spec definition. 

"10.4.5 404 Not Found
The server has not found anything matching the Request-URI. No indication is given of whether the condition is temporary or permanent. The 410 (Gone) status code SHOULD be used if the server knows, through some internally configurable mechanism, that an old resource is permanently unavailable and has no forwarding address. This status code is commonly used when the server does not wish to reveal exactly why the request has been refused, or when no other response is applicable."

In this case, I understand that we could return 204. 

Reference: https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

davidffrench: You have sparked an age-old debate which there never seems to be a definitive answer on. Here are some resources on the matter.

- https://stackoverflow.com/questions/11746894/what-is-the-proper-rest-response-code-for-a-valid-request-but-an-empty-data
- https://softwareengineering.stackexchange.com/questions/203492/when-to-use-http-status-code-404-in-an-api

Here is how I see it for our use case. We should probably be returning a 204 on /apps if there are no apps that are yet bound. We should be returning a 404 on /apps/{id} if that specific app does not exist.

davidffrench: Changed the apps http handler to return a 204 now.

camilamacedo86: Hi @davidffrench,

IHMO the nomenclature `NoContent` will be more appropriate too.

davidffrench: Not for a generic error message in my opinion, NotFound is a much more widely known message. The purpose of the errors model is to have no relation to the http delivery layer. NotFound is more generic in my opinion.

camilamacedo86: The point is that here will return 404 when a data does not exist in the database. For it returns the code 204 according to HTTP code definitions then is required to add [here](https://github.com/aerogear/mobile-security-service/blob/ecb03c5d1f8dfb515f844b85897ac04410a8cc69/pkg/httperrors/httperrors.go#L68) and the func called will be NoContent.

davidffrench: @camilamacedo86 I completely agree that for the /apps endpoint a 204 is what makes the most sense to return. This is not the case for /apps/{id} which is a specific resource. As defined in the RFC for HTTP 1.1 (https://tools.ietf.org/html/rfc7231#page-59)

>   The 404 (Not Found) status code indicates that the origin server did
   not find a current representation for the target resource or is not
   willing to disclose that one exists.

A 204 (https://tools.ietf.org/html/rfc7231#section-6.3.5) could be interpreted to be used in this case but it should not be.

camilamacedo86: HI @davidffrench,
 
Note that `current representation for the target resource and is not willing to disclose that one exists` is referent to `indicates that the origin server`. The target is the web server which means the endpoint. However, in the case of `/apps/{id}` it definitely could be applied since the ID is part of the endpoint then it also could mean that the endpoint does not exist. 

So, anyway, should not it here also have the 204 mapped since it is a generic http erros mapping?

pkg/web/apps/apps_service.go
```diff
@@ -29,7 +28,7 @@ func (a *appsService) GetApps(c echo.Context) (*[]models.App, error) {
 	apps, err := a.psqlRepository.GetApps(c)
 
 	if err != nil {
-		return nil, httperrors.NotFound(c, "No apps found")
+		return nil, models.ErrNotFound
```
camilamacedo86: Is not this the case that we in common sense agreed that should return the 204? Is not here the reference to the endpoint /apps/?

camilamacedo86:
Hi @davidffrench, 

Great work 🥇. I understand that it is missing just the change regards the 204 HTTP error response to be returned in the endpoint /apps and all is fine 👍 .

pkg/httperrors/httperrors.go
```diff
@@ -122,3 +123,22 @@ func HTTPError(c echo.Context, statusCode int, message string) (e error) {
...
+	case models.ErrConflict:
+		return Conflict(c, err.Error())
+	case models.ErrBadParamInput:
+		return BadRequest(c, err.Error())
+	case models.ErrUnauthorized:
+		return Unauthorized(c, err.Error())
+	default:
+		return InternalServerError(c, "")
```
camilamacedo86: Should not be better here return the error message instead of "" ? If we keep in this way, it will return Internal Server Error 500 without any information which could be helpful to understand the issue. Wdyt?

pkg/httperrors/httperrors.go
davidffrench: Thanks, Camila, I understand it could be interpreted as that. Our API will not be exposed for public use, we will not need to worry about this scenario.

pkg/httperrors/httperrors.go
davidffrench: It will end up returning a message. If a message is not supplied [1], it will use the message from the status code map[2]. I did not want to duplicate the text and have to pass the string argument.

[1] https://github.com/aerogear/mobile-security-service/blob/4e7f7c8a890edad98e35c2d07bf4bde64ac9815d/pkg/httperrors/httperrors.go#L115
[2] https://github.com/aerogear/mobile-security-service/blob/4e7f7c8a890edad98e35c2d07bf4bde64ac9815d/pkg/httperrors/httperrors.go#L39

camilamacedo86: However, should we not return something as err.getMessage() here instead of "" in the following line? 

```nodejs
return InternalServerError(c, "")
```
I mean, for example, the issue is with the db connection, so the error message will be something as `connection refuse`. Another example could be something like `number exception format`. We could return this message to clarify it in order to be easier to troubleshooting it. Wdyt?

pkg/httperrors/httperrors_test.go
```diff
@@ -41,3 +43,57 @@ func TestHTTPError(t *testing.T) {
 		})
 	}
 }
+
+func TestGetHTTPResponseFromErr(t *testing.T) {
```
craicoverflow: The function name is quite verbose. It leads to a long function name:

```go
return httperrors.GetHTTPResponseFromErr(c, err)
```

Any way this could be slimmed down so it looks like:

```go
return httperrors.GetError(c, err)
```

pkg/web/apps/apps_http_handler.go
```diff
@@ -26,8 +29,13 @@ func NewHTTPHandler(e *echo.Echo, s Service) *HTTPHandler {
 func (a *HTTPHandler) GetApps(c echo.Context) error {
 	apps, err := a.Service.GetApps(c)
 
+	// If no apps have been found, return a HTTP Status code of 204 with no response body
+	if err == models.ErrNotFound {
+		return c.NoContent(http.StatusNoContent)
```
craicoverflow: I know it's an echo function not something you made, but it seems counter-intuitive to to have a specific function for http status "No Content", yet you still have to provide the status code :thinking:

pkg/httperrors/httperrors_test.go
austincunningham: In favor of longer method names as more descriptive

austincunningham: happy with long function names. As projects grow the more descriptive the method and function the better

pkg/httperrors/httperrors.go
davidffrench: Line 142 is just the default in the switch statement, if we have called this function with an unknown model error. This should never happen. But as the default, I am returning a 500, just so there is a default.

pkg/httperrors/httperrors_test.go
davidffrench: I am also in favour of longer descriptive function names.

pkg/web/apps/apps_http_handler.go
davidffrench: Ah, so it is not for the status code. it is `Context#NoContent(code int) can be used to send empty body with status code.`

PR #35, opened 2019-02-08
feat(database): Create PostgreSQL database (AEROGEAR-8384)

craicoverflow:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-8384

## What

* Created Postgres database based on ERD
* Implemented server code to connect to database
* Implemented code to create tables
* Refactored unit tests to use database connection
* Changed App ID data type
* Updated README with database connection variables

## Why

The service needs a persistent database to store app information. The API will be used to save and retrieve information from that database.

## How

* Used [ERD](https://user-images.githubusercontent.com/1596014/52293332-7229e380-296e-11e9-9bc4-5322f0fbe0ab.png) to create the database.
* Added code in Go to connect to database using `database/sql` and `[lib/pq](https://github.com/lib/pq)`

## Verification Steps
 
1. Run the database container using `docker-compose up -d`. This will create the database
2. Run the API using `go run cmd/mobile-security-service/main.go`. This will create the tables if they don't already exist.
3. Go to an endpoint, such as http://localhost:3000/api/apps
4. If the server runs successfully and you can hit an endpoint, the database is connected.
5. Run integration tests: `make test-integration`
6. Run unit tests: `make test`

## Checklist:

## Progress

- [ ] Finished task

## Additional Notes

The logic in the repository does not yet get the information from the tables, but the infrastructure is now in place for us to do this. This can be done in a separate PR.


davidffrench:
@craicoverflow I am having some issues with the verification steps.

If I run `docker compose up -d` and go to `http://localhost:3000/api/apps`, I get a response of not found, which is to be expected.

If I run `docker compose up db` and then run `go run cmd/mobile-security-service/main.go`, I get the following error `panic: failed to connect to SQL database: pq: database "mobile_security_service" does not exist`

craicoverflow:
> @craicoverflow I am having some issues with the verification steps.
> 
> If I run `docker compose up -d` and go to `http://localhost:3000/api/apps`, I get a response of not found, which is to be expected.
> 
> If I run `docker compose up db` and then run `go run cmd/mobile-security-service/main.go`, I get the following error `panic: failed to connect to SQL database: pq: database "mobile_security_service" does not exist`

@davidffrench it's possible that the database container had not finished creating the database when you ran `go run cmd/mobile-security-service/main.go`. Could you try again? I have ran both these scenarios and they worked.

Also, you should not get a response of not found for  `http://localhost:3000/api/apps` as the repository function does not retrieve this data from the database yet. This will be in a different task/PR as it is out of the scope of this task.

davidffrench:
@craicoverflow Ignore the above comment. I figured out what was wrong after some debugging. I still have a previous docker container which had the db listed as aerogear_mobile_security_service. I figured this out after running 

1. `docker exec -it e4e4e693f8f7 bash`
2. `psql -U postgres`
3. `\l`

After removing the docker container, everything is now working as expected. Just in case anyone else has the same problem, I will leave this info here.

README.md
```diff
@@ -85,9 +85,28 @@ Now the application will use the values defined in `.env`.
...
+| PGAPPNAME         |                         | The [application_name](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNECT-APPLICATION-NAME) connection parameter |
+| PGSSLCERT         |                         | The [sslcert](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNECT-SSLCERT) connection parameter.                  |
+| PGSSLKEY          |                         | The [sslkey](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNECT-SSLKEY) connection parameter.                    |
+| PGSSLROOTCERT     |                         | The [sslrootcert](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT) connection parameter           |
+
 #### Database Entity Relationship Diagram
+
 <img src="https://user-images.githubusercontent.com/1596014/52293332-7229e380-296e-11e9-9bc4-5322f0fbe0ab.png" width="700" height="300">
 
 *Lucid Chart Link:* https://www.lucidchart.com/documents/edit/6293d791-8f70-47bb-8136-438128c27129/0
```
camilamacedo86: Better we merge this one : https://github.com/aerogear/mobile-security-service then update this point here

cmd/mobile-security-service/main.go
```diff
@@ -56,12 +57,23 @@ func initLogger(level, format string) {
 
 // Invoke handlers, services and repositories here
 func initHandlers(e *echo.Echo, c config.Config) {
+
 	// Prefix api routes
 	apiRoutePrefix := c.ApiRoutePrefix
 	apiGroup := e.Group(apiRoutePrefix)
```
camilamacedo86: The handler lawyer is responsible to receive the requests. The lawyer which will connect to the database is the repository one. So, I am not sure about this code here. I mean I don't think it should be imported in this lawyer at all.

craicoverflow: @camilamacedo86 what is a lawyer? I'm not familiar with this term in this context.

Perhaps this function should not be called `initHandlers` as you are right, it is doing more than this, it is initialising the service and repository as well.

README.md
craicoverflow: I think you're Adoc PR will be merged before this so I will probably have to pull in those changes and modify this before merging anyway :smile:

cmd/mobile-security-service/main.go
davidffrench: I would be inclined to split the db connect to it's own function, called from main before echo is instansiated and pass it through to this function.

internal/db/db.go
```diff
@@ -0,0 +1,66 @@
...
+// the necessary tables for the API (if they don't exist).
+func Setup(db *sql.DB) error {
+	if db == nil {
+		return errors.New("cannot setup database, must call Connect() first")
+	}
+
+	if _, err := db.Exec(
+		`
+		CREATE TABLE IF NOT EXISTS app (
+	    	id uuid NOT NULL UNIQUE,
```
davidffrench: Minor indentation change here

cmd/mobile-security-service/main.go
damienomurchu: @craicoverflow I think @camilamacedo86 means `layer`

+1 on `initHandlers()` not being the best place for db setup, perhaps move the db setup either to `main()` or its own dedicated method that would be called in `main()`, or else rename `initHandlers` to reflect what it's now doing?

pkg/web/apps/apps_http_handler_test.go
```diff
@@ -6,29 +6,39 @@ import (
...
+
 	"github.com/labstack/echo"
 )
 
-var mockPostgresRepository = NewPostgreSQLRepository()
-var mockService = NewService(mockPostgresRepository)
-
 func TestHTTPHandler_GetApps(t *testing.T) {
+	config := config.Get()
+	dbConn, err := db.Connect(config.DB.ConnectionString, config.DB.MaxConnections)
```
davidffrench: This makes our test dependant on the db running. We need to avoid this in our unit tests. The other mobile services split it so their integration tests have a dependency on the db but not unit tests.

internal/db/db.go
damienomurchu: `gofmt -w internal/db/db.go` should take care of that :+1:

cmd/mobile-security-service/main.go
camilamacedo86: Hi @craicoverflow,

`Layer` it is regarding the architecture of the project. Each layer has one kind of responsibility. 

So, the request will go through them as follows. 

https://cdn-images-1.medium.com/max/1600/1*0bgo3z8DsVR0WbEA-D6FEg.png

In this way, the layer responsible to connect to the database and get the data "Data Access Object" here is the repository

pkg/web/apps/apps_http_handler_test.go
craicoverflow: @davidffrench I have added an integration tag to the handler, repository and service test files so that they will only when the integration tag is supplied during `go test`. Is this right?

davidffrench: Yes, that should do it but wasn't for me locally.

davidffrench: As discussed in person, integration tests should be separated out.

cmd/mobile-security-service/main.go
craicoverflow: > Hi @craicoverflow,
> 
> `Layer` it is regarding the architecture of the project. Each layer has one kind of responsibility. So, the request will go through them as follows.
> 
> 1_0bgo3z8dsvr0wbea-d6feg
> 
> In this way, the layer responsible to connect to the database and get the data "Data Access Object" here is the repository. The handlers should not call the db directly.

Oh _layer_! You had said _lawyer_ which really confused me :grin:

pkg/web/apps/apps_http_handler_test.go
davidffrench: @craicoverflow Apologies, I see the changes for the integration tests have split everything out. This is not exactly what I meant. There should be a single file for the integration api tests just hitting the endpoints with the database setup. Exactly like https://github.com/aerogear/aerogear-app-metrics/blob/master/pkg/test/metrics-api_test.go have it setup. This will cover end to end integration of the components working together.

The db_test.go file can stay where it was but just with the `// +build integration` comment.

The HTTP handler, service and repository files still need there own unit test files associated with them. 

Let me know if you have any questions. If you prefer, catch me tomorrow and we can collaborate.

craicoverflow: :+1:

cmd/mobile-security-service/main.go
camilamacedo86: The Grammar app sometimes leads me to some misunderstands. Sorry, for that.

cmd/mobile-security-service/main.go
```diff
@@ -54,14 +58,30 @@ func initLogger(level, format string) {
...
+	if err := db.Setup(dbConn); err != nil {
+		panic("failed to perform database setup: " + err.Error())
+	}
+
+	return dbConn
+}
+
 // Invoke handlers, services and repositories here
-func initHandlers(e *echo.Echo, c config.Config) {
+func initHandlers(e *echo.Echo, c config.Config, dbConn *sql.DB) {
```
camilamacedo86: It shows still not make too much sense. Why a func with the name initHandler is using the database?

pkg/config/config.go
```diff
@@ -32,7 +39,44 @@ func Get() Config {
...
+// builds a libpq compatible connection string e.g. "user=postgresql host=localhost password=postgres
+func getDBConnectionString() string {
+
+	// These are all of the options supported by pq
+	dbConnectionConfig := map[string]string{
+		"dbname":                    getEnv("PGDATABASE", "mobile_security_service"),
+		"user":                      getEnv("PGUSER", "postgresql"),
+		"password":                  getEnv("PGPASSWORD", "postgres"),
+		"host":                      getEnv("PGHOST", "localhost"),
+		"port":                      getEnv("PGPORT", "5432"),
```
camilamacedo86: Should it not be env properties? Will ii always be a HOST=localhost?

pkg/db/db.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Setup uses the existing database connection to create
+// the necessary tables for the API (if they don't exist).
+func Setup(db *sql.DB) error {
+	if db == nil {
+		return errors.New("cannot setup database, must call Connect() first")
+	}
+
+	if _, err := db.Exec(`
+		CREATE TABLE IF NOT EXISTS app (
```
camilamacedo86: Should it not better if it is in a seeds script called in the make setup for example?

pkg/test/apps_http_handler_test.go
```diff
@@ -0,0 +1,42 @@
+// +build integration
```
camilamacedo86: The dir structure here shows not to be properly organized. Here you add it in a test dir, however, you also add tests on  `pkg/web/apps/apps_psql_repository_test.go`. 

IHMO we should organize it by layers and not by business model. For example:
/handler
/service
/repository

Following the main reasons:

1. It makes clear for the devs the architecture defined to the project
2. Ita void confusion and help the dev not to do implementation which thoughts the layers
3. It makes easier the understand and where we should looking for or add the files

Also, shows that in go the test usually stays with the implementations. So, for example, the file app_handler_test will be together with app_handler and both in the dir /handler if my suggestion is accepted by the team.

cmd/mobile-security-service/main.go
damienomurchu: +1 `initHandlers` badly named at the least now

pkg/config/config.go
craicoverflow: These are env properties. localhost is the default value but they can be overridden with environment variables. Refer to the [documentation](https://github.com/aerogear/mobile-security-service#add-your-own-env-file) on using your environment variables.

pkg/db/db.go
craicoverflow: This was the decision agreed upon as it is used in app-metrics. Ideally long term we should have migrations in place.

pkg/test/apps_http_handler_test.go
craicoverflow: @camilamacedo86 this is outside of the scope of this PR. We have already agreed that the current approach is the best one.

[Trying Clean Architecture on Golang](https://hackernoon.com/golang-clean-archithecture-efd6d7c43047)

cmd/mobile-security-service/main.go
craicoverflow: Agreed that the function name is bad but this would probably require change in a separate PR. I think it could all be refactored into a function called `setup` or something similar.

@camilamacedo86 the reason `initHandlers` is using the database is because the handlers require the database to be injected into them.

pkg/test/apps_http_handler_test.go
camilamacedo86: @craicoverflow the project shows not following what has in the Blog Post added for you too. See it here https://hackernoon.com/trying-clean-architecture-on-golang-2-44d615bf8fdf. 

screenshot 2019-02-18 at 16 10 36

IHMO: We should create a task to skill up it and review, recheck if it is all place, if not them do the changes and add the argumentation in the PR for we are able to discuss it. Wdyt @davidffrench?

davidffrench: There is no need for the sub folders until we have more repositorys and delivery methods in my opinion. It just makes the folder structure more verbose in my opinion. It is unlikely we will have any additional storage or delivery methods in the future either.

craicoverflow: @camilamacedo86 this is not related to this PR. This can be discussed in our other communication channels if needed. We are not following the format from the blog post strictly, as you can see we are also following the [go-standard-layout](https://github.com/golang-standards/project-layout).

The main difference between ours and the blog post is they have subfolders per package. This is a valid discussion but it won't get the appropriate attention here.

camilamacedo86: @davidffrench the main diff over what has in the blog post and it this project is not the subfolders. According to the blog shows that it should be something as  

/apps
|----repository
/versions
|----repository
/devices
|----repository
/middleware
---> here will be the routers 

instead of the pkg.

davidffrench: @camilamacedo86 I get you. As Enda said above, we are using a combination of both clean architecture and go-standard-layout. The clean architecture from the blog post was one person's interpretation of a language agnostic clean architecture solution.

craicoverflow: @davidffrench +1. 

We are also following the **aerogear format** - see [app-metrics](https://github.com/aerogear/aerogear-app-metrics). This also follows [go-standard-layout](https://github.com/golang-standards/project-layout).

PR #36, opened 2019-02-11
Readme asciidoc

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8377

## What
 1) README -> asscidoc 
2) package name 
3) .gitignore improvements

## Why
Feedback from https://github.com/aerogear/mobile-security-service/pull/31

## Verification Steps
Readme should render correctly
 
## Checklist:


ui/README.adoc
```diff
@@ -0,0 +1,17 @@
+= Mobile Security Service UI
```
camilamacedo86: I'd like to suggest to add on top of this file the following code. It will keep the same style from the other readmes. 

```
ifdef::env-github[]
:status:
:outfilesuffix: .adoc
:caution-caption: :fire:
:important-caption: :exclamation:
:note-caption: :paperclip:
:tip-caption: :bulb:
:warning-caption: :warning:
:table-caption!:
endif::[]

:toc:
:toc-placement!:
```

ui/README.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+== Local Development using VSCode Chrome Debugger
+
```
camilamacedo86: Should not better add here `=== Getting started` on top of `* Run:`? Wdyt?

ui/README.adoc
```diff
@@ -0,0 +1,17 @@
...
+* Run:
+[source,shell]
+----
+npm start
+----
+to run the app in the development mode on localhost:3000
```
camilamacedo86: Wdyt about it be something as a *NOTE* : It will start the app in the local development as `localhost:3000`?

ui/README.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+* Open the project in VSCode
```
camilamacedo86: Wdyt it be a section as `=== Open the project in VSCode`?

ui/README.adoc
```diff
@@ -0,0 +1,17 @@
...
+* Run in debug mode (config is provided)
```
camilamacedo86: Wdyt it be a section as `=== Debuging the project in VSCode`?

ui/README.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+Find more info on VSCode chrome debugger https://code.visualstudio.com/docs/nodejs/reactjs-tutorial#_debugging-react[here].
```
camilamacedo86: Wdyt about it here be a *TIP*?

ui/README.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+This project was bootstrapped with https://github.com/facebook/create-react-app[Create React App].
```
camilamacedo86: Wdyt about it be a *NOTE* ?

ui/README.adoc
camilamacedo86: Following an image in order to clarifies : 

screenshot 2019-02-11 at 12 13 24

ui/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "ui",
+  "name": "mobile-security-service-ui",
   "version": "0.1.0",
```
camilamacedo86: Why the version is 0.1.0 and not 0.0.1?

laurafitzgerald: It's the default version which was provided by `create-react-app`. I didn't update but I can. There is a versioning policy for aerogear mobile-services projects. I will share it in chat.

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
camilamacedo86: Should we commit it? 

IHMO it is just the VSCODE file created for who uses this IDEA then it is not pertinent to the project and people are able to use other IDEAs. I don't think that it should be committed.

ui/package.json
damienomurchu: Initial versioning can be a bit confusing. Below from https://semver.org/:
```
How should I deal with revisions in the 0.y.z initial development phase?

The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.
```

.vscode/launch.json
laurafitzgerald: @camilamacedo86 I also wasn't using vscode but I was advised by @davidffrench that the team all use it so I installed it and I wrote the steps to develop/debug locally based on the use of that. In that case it could be helpful to have the configuration already in the project files in order to help someone new to the project, that way they don't need to do the steps to generate it.

ui/README.adoc
laurafitzgerald: Will do @camilamacedo86 I didn't include the toc as it was only one entry and the README is very short. I don't think it's necessary as I believe this README is going to be very short anyway as I don't want to duplicate content from the root README

ui/README.adoc
laurafitzgerald: I tried it with the heading but it looks a little funny without any content inside of it. I updated it to be an ordered list.

ui/README.adoc
laurafitzgerald: I tried it with the heading but it looks a little funny without any content inside of it. I updated it to be an ordered list.

ui/README.adoc
laurafitzgerald: I reordered the text and the code rather than adding a note or tip. I think it looks a little better. wdyt?

.vscode/launch.json
camilamacedo86: @laurafitzgerald it is fine. However, this file will be generated automatically when open the project in the IDEA VSCode. So, it is not required to commit it. I understand that we should not commit the IDEA config files and they should be added in the .gitignore. 

PS.: I do not use VSCode. I normally use the Intellij ideas :-)

austincunningham: disagree automatic launch.json is very different
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch",
            "type": "go",
            "request": "launch",
            "mode": "auto",
            "program": "${fileDirname}",
            "env": {},
            "args": []
        }
    ]
}
```

laurafitzgerald: @austincunningham are you saying we should include it?

craicoverflow: @laurafitzgerald he is.

austincunningham: :+1:

camilamacedo86: IHMO it can be added in the readme as config for who would like to use VSCode IDEA since it has customizations and it is not the default file generated. **REASON: It is not a good practice to add local config files in the projects. It is specific for who would like to use VS CODE and it and is not part of the project at all.** 

Also, note that:

screenshot 2019-02-12 at 09 31 16

Following some links.  

- https://softwareengineering.stackexchange.com/questions/202051/what-should-i-include-in-my-repositiory-from-ide-projects

> The repo should be clean and tidy and only have those things which are relevant to the project not your environment.

- https://hackernoon.com/exclude-files-from-git-without-committing-changes-to-gitignore-986fa712e78d

> Maybe there are some files, however, that you don’t want to commit to version control because:
> 
> Local settings, scripts or other configuration files that are specific to your local development environment.

- https://gist.github.com/pandeiro/1552496#dont

screenshot 2019-02-12 at 09 18 20

- https://datree.io/blog/top-10-github-best-practices/

> We strongly recommend against committing your local config files to version control.

damienomurchu: Tend to agree with @camilamacedo86 as it's IDE-specific. An alternative to including the file with the repo would be to document it in the README

laurafitzgerald: @camilamacedo86 @damienomurchu yes the more I think about it the more I think we shouldn't include it.  I will make the required updates to the README to provide guidance on how to get the config set up locally.

camilamacedo86: Hi @laurafitzgerald you made a great job here. I am approving it since IHMO all here shows perfect now !!!.

laurafitzgerald:
@camilamacedo86 thanks. I see one problem with the README. I will fix it before I merge it.

PR #38, opened 2019-02-11
adding eslint

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8505

#### Commit 1: [adding eslint](https://github.com/aerogear/mobile-security-service/pull/38/commits/d33f26035a2f9a835eab4743dcac6309ebaa34ca)
Add ESLint dependencies
Add config file
Fix some linting problems

#### Commit 2: [adding-standard-react](https://github.com/aerogear/mobile-security-service/pull/38/commits/d49acc7a097397bd0e161d4ef887359f4f5d7829)
Adding more specific standard-react config and related dependencies
See more here: https://github.com/standard/eslint-config-standard-react

## Why
See JIRA

## How
Following: https://code.visualstudio.com/docs/nodejs/reactjs-tutorial#_linting
I installed VSCode ESLint plugin and ran the comman eslint: Create ESLint configuration

## Verification Steps
Make some changes which could trigger a linting warning

## Checklist:

## Progress

- [x] Added Initial setup
- [ ] TODO: Agree as a team if we want to have alternative linting standards to the default options provided by ESLINT

which are the following:

image

image

What I've pushed up is a very basic setup with the https://github.com/standard/standard config enabled. It's possible there might have been some discussion about what we want to go with as a team. Very much open to debate here but would be equally happy to not reinvent the wheel if there are existing standards we could leverage.

**EDIT**: Second commit has further options See **What** section above.

### DIFF
--- a/ui/.eslintrc.json
+++ b/ui/.eslintrc.json
@@ -1,0 +1,7 @@
+{
+    "extends": ["react-app","standard"],
+    "rules": {
+        "semi": [2, "always"],
+        "prefer-const": [2]
+    }
+}
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -81,6 +81,18 @@
         "@babel/types": "^7.0.0"
       }
     },
+    "@babel/helper-create-class-features-plugin": {
+      "version": "7.3.2",
+      "resolved": "https://registry.npmjs.org/@babel/helper-create-class-features-plugin/-/helper-create-class-features-plugin-7.3.2.tgz",
+      "integrity": "sha512-tdW8+V8ceh2US4GsYdNVNoohq5uVwOf9k6krjwW4E1lINcHgttnWcNqgdoessn12dAy8QkbezlbQh2nXISNY+A==",
+      "requires": {
+        "@babel/helper-function-name": "^7.1.0",
+        "@babel/helper-member-expression-to-functions": "^7.0.0",
+        "@babel/helper-optimise-call-expression": "^7.0.0",
+        "@babel/helper-plugin-utils": "^7.0.0",
+        "@babel/helper-replace-supers": "^7.2.3"
+      }
+    },
     "@babel/helper-define-map": {
       "version": "7.1.0",
       "resolved": "https://registry.npmjs.org/@babel/helper-define-map/-/helper-define-map-7.1.0.tgz",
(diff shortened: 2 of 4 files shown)


laurafitzgerald:
> Thanks for doing this @laurafitzgerald . I have no preference on which eslint config to use. As I have never done a React production application before, I do not know the pros/cons of the different options. I am happy to go with the AirBnB config.

Hi @davidffrench The image might be misleading, I chose the Standard one (second option)

damienomurchu:
@laurafitzgerald +1 for linting. What do you think about aligning with what is being used by the other aerogear repos? MDC uses [this set of eslint rules](https://github.com/aerogear/mobile-developer-console/blob/master/ui/.eslintrc.json) (not sure what the other projects use).

davidffrench:
@laurafitzgerald It might be worth kicking off a mail thread to aerogear to get a consensus. I agree with Damien, staying consistent with existing Mobile Services is a better option for all developers to easily switch between projects.

ui/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+    "extends": ["standard", "standard-react"]
+}
```
camilamacedo86: In the RHMAP we had as default use the rules of ngui. Is just this rule enough? Should we apply the same rules here?

damienomurchu: @camilamacedo86 maybe/ maybe not. The conventions the js community favour may have moved on, but also ngui was backbone, and we will need to include react-specific eslint setting in our config

laurafitzgerald:
Thanks for the feedback/comments guys. I'll kick off a thread on aerogear to see if it makes sense to use the same setup as MDC. I think it makes sense to have consistency across the coding standards for each of the repos.

ui/.eslintrc.json
camilamacedo86: It is fine. The idea of this comment is just for we check: Is just this rule enough for this project? Have someone standard that we should follow here? What rules has in the others aerogear projects?

laurafitzgerald:
Consensus [1] already seems to be to use the same one from MDC, will go ahead and update the config

[1] https://groups.google.com/forum/#!topic/aerogear/nZxXj5Uu8ig

wei-lee:
> Consensus [1] already seems to be to use the same one from MDC, will go ahead and update the config
> 
> [1] https://groups.google.com/forum/#!topic/aerogear/nZxXj5Uu8ig

@laurafitzgerald I am not sure that's what have been agreed in the thread, at least it's not what I meant? I meant that we stick with the standard style that you have in this one, but maybe considering allow semicolons.

laurafitzgerald:
> > Consensus [1] already seems to be to use the same one from MDC, will go ahead and update the config
> > [1] https://groups.google.com/forum/#!topic/aerogear/nZxXj5Uu8ig
> 
> @laurafitzgerald I am not sure that's what have been agreed in the thread, at least it's not what I meant? I meant that we stick with the standard style that you have in this one, but maybe considering allow semicolons.

@wei-lee sorry about that, I misread your mail. Will progress with what you said. Using standard and adding the semi-colon.

laurafitzgerald:
@wei-lee @davidffrench @damienomurchu @camilamacedo86 pr is now updated to use semistandard. 

Test using `npm run lint` 

You should see no linting errors

Remove semicolon and re-run
You will get errors.

craicoverflow:
I think we should use [prefer-const](https://eslint.org/docs/rules/prefer-const) rule. It ensures we use a const where the variable is not changed after its original definition.

damienomurchu:
I think we should also use the [strict](https://eslint.org/docs/rules/strict) rule to enforce `use strict`

laurafitzgerald:
> I think we should also use the [strict](https://eslint.org/docs/rules/strict) rule to enforce `use strict`

Hi @damienomurchu 
Happy to add this in. I did some digging on it though as I wasn't sure what it would give us. What I found is outlined below.

`Module code is always strict mode code.` at least in ES6 [1]
and
`both export and import declaration make it a module`. 

I could be parsing it wrong but I think that means it's implied and I'm also seeing errors if I try to add "use strict" to any of the existing React Code.

`'use strict' is unnecessary inside of modules `

There is a React Strict Mode [2] but I think that's a separate thing.

[1] https://stackoverflow.com/questions/29283935/which-ecmascript-6-features-imply-strict-mode
[2] https://stackoverflow.com/questions/53183362/what-is-strictmode-in-react

PR #39, opened 2019-02-11
fix(apps): Decouple Postgres from service layer (AEROGEAR-8522)

davidffrench:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8522
The Postgres repository interface is tightly coupled to our service layer, this should not be the case. 

## What
The repository layer interface needs to be abstracted away from a specific database implementation. 

## Why
Clean separation of concerns

## How
Our Postgres repository file will implement the generic repository interface which is also used in our service.

## Verification Steps
None required

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/web/apps/apps_repository.go
+++ b/pkg/web/apps/apps_repository.go
@@ -1,0 +1,11 @@
+package apps
+
+import (
+	"github.com/aerogear/mobile-security-service/pkg/models"
+	"github.com/labstack/echo"
+)
+
+// Repository represent the app's repository contract
+type Repository interface {
+	GetApps(c echo.Context) (*[]models.App, error)
+}
--- a/pkg/web/apps/apps_repository.go
+++ b/pkg/web/apps/apps_repository.go
@@ -2,10 +2,9 @@ package apps
 
 import (
 	"github.com/aerogear/mobile-security-service/pkg/models"
-	"github.com/labstack/echo"
 )
 
 // Repository represent the app's repository contract
 type Repository interface {
-	GetApps(c echo.Context) (*[]models.App, error)
+	GetApps() (*[]models.App, error)
 }
--- a/pkg/web/apps/apps_service.go
+++ b/pkg/web/apps/apps_service.go
@@ -2,13 +2,12 @@ package apps
 
 import (
 	"github.com/aerogear/mobile-security-service/pkg/models"
-	"github.com/labstack/echo"
 )
 
 type (
 	// Service defines the interface methods to be used
 	Service interface {
-		GetApps(ctx echo.Context) (*[]models.App, error)
+		GetApps() (*[]models.App, error)
 	}
 
 	appsService struct {
@@ -24,8 +23,8 @@ func NewService(repository Repository) Service {
 }
 
 // GetApps retrieves the list of apps from the repository
-func (a *appsService) GetApps(c echo.Context) (*[]models.App, error) {
-	apps, err := a.repository.GetApps(c)
+func (a *appsService) GetApps() (*[]models.App, error) {
+	apps, err := a.repository.GetApps()
 
 	if err != nil {
 		return nil, models.ErrNotFound
--- a/pkg/web/apps/apps_service.go
+++ b/pkg/web/apps/apps_service.go
@@ -12,20 +12,20 @@ type (
 	}
 
 	appsService struct {
-		psqlRepository PostgreSQLRepository
+		repository Repository
 	}
 )
 
 // NewService instantiates this service
-func NewService(psqlRepository PostgreSQLRepository) Service {
+func NewService(repository Repository) Service {
 	return &appsService{
-		psqlRepository: psqlRepository,
+		repository: repository,
 	}
 }
 
 // GetApps retrieves the list of apps from the repository
 func (a *appsService) GetApps(c echo.Context) (*[]models.App, error) {
-	apps, err := a.psqlRepository.GetApps(c)
+	apps, err := a.repository.GetApps(c)
 
 	if err != nil {
 		return nil, models.ErrNotFound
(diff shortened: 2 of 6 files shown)


pkg/web/apps/apps_repository.go
```diff
@@ -0,0 +1,11 @@
+package apps
+
+import (
+	"github.com/aerogear/mobile-security-service/pkg/models"
+	"github.com/labstack/echo"
+)
+
+// Repository represent the app's repository contract
+type Repository interface {
+	GetApps(c echo.Context) (*[]models.App, error)
```
wei-lee: the request context should not be passed to the `Repository` layer. You should get the information you need (business object) for the request at the web layer, and from there on, you shouldn't use the request context anywhere else.

pkg/web/apps/apps_service.go
```diff
@@ -12,20 +12,20 @@ type (
...
-func NewService(psqlRepository PostgreSQLRepository) Service {
+func NewService(repository Repository) Service {
 	return &appsService{
-		psqlRepository: psqlRepository,
+		repository: repository,
 	}
 }
 
 // GetApps retrieves the list of apps from the repository
 func (a *appsService) GetApps(c echo.Context) (*[]models.App, error) {
```
wei-lee: same here. The server layer should really need to know about the web layer.

pkg/web/apps/apps_repository.go
davidffrench: @wei-lee Thank you Wei, we are aware of this and this is gettting changed in Austins PR here - https://github.com/aerogear/mobile-security-service/blob/473aaadfe303f0a2a7eb776cdb1797fdb9368ceb/pkg/web/apps/apps_service.go#L40

davidffrench: Thanks @wei-lee . This was being tackled in a separate PR but after talking to Austin, I will make the change here

davidffrench: Changed now

pkg/web/apps/apps_service.go
davidffrench: Changed now

pkg/web/apps/apps_service.go
```diff
@@ -2,30 +2,29 @@ package apps
...
 	return &appsService{
-		psqlRepository: psqlRepository,
+		repository: repository,
 	}
 }
 
 // GetApps retrieves the list of apps from the repository
-func (a *appsService) GetApps(c echo.Context) (*[]models.App, error) {
-	apps, err := a.psqlRepository.GetApps(c)
+func (a *appsService) GetApps() (*[]models.App, error) {
```
laurafitzgerald: Was the context object being used in this function somewhere before this change?

davidffrench: it was not being used anywhere. If we wanted too, we could strip out the Go context from the echo Context to pass down to our future database calls. After talking to Craig Brookes, this is not needed.

craicoverflow:
> It shows perfect for me.  Great Job @davidffrench 
> 
> Also, I'd like we create a task to define the dirs of the project by the layers and not the business model. IHMO: It makes easier for the dev do not through the layers and understand the architecture. For example
> 
> /handler
> /service
> /repository

@camilamacedo86 it is common practice in Go to organise by responsibility - [source](https://rakyll.org/style-packages/):

## Organize by responsibility

A common practise from other languages is to organize types together in a package called models or types. In Go, we organize code by their functional responsibilities.

```go
package models // DON'T DO IT!!!
```

```go
// User represents a user in the system.
type User struct {...}
```

Rather than creating a models package and declare all entity types there, a User type should live in a service-layer package.

```go
package mngtservice

// User represents a user in the system.
type User struct {...}

func UsersByQuery(ctx context.Context, q *Query) ([]*User, *Iterator, error)

func UserIDByEmail(ctx context.Context, email string) (int64, error)
```

camilamacedo86:
@craicoverflow, 

Regards: 


So according to your above comment, we are not doing it properly. Since that what you added here over what we should NOT do is exactly what is done. See the following image.

screenshot 2019-02-19 at 10 18 25

 c/c @davidffrench I think it is part of my comment here: https://github.com/aerogear/mobile-security-service/pull/35#discussion_r. Wdyt over we create a task as suggested in this comment?

craicoverflow:
@camilamacedo86 we had to do it for the models as there is no handler, service, repository for the version model.

davidffrench:
@camilamacedo86 I am not sure exactly what your ask is here? 

This PR that was merged a week ago is probably not the best place for this conversation.

camilamacedo86:
@craicoverflow my main point here was exactly it each part of the project and file be following a different rule which makes it very confusing.

I am not sure if "we had to do it for the models as there is no handler, service, a repository for the version model." it justify this need. 

@davidffrench my suggestion was to create a task, skill-up and then create a PR with the changes if required in order to provide the discussion and we be able to improve it. I am not sure where could we discuss this subject. However, I believe that with this comment and the other one [here]( https://github.com/aerogear/mobile-security-service/pull/35#discussion_r) I could clarify my idea. Please, feel free to move forward without a task if you don't see the need or let me know if you think that it should be a good approach.

davidffrench:
@camilamacedo86 I don't believe the current project structure is hard to read but I have been working on it for a little while so it makes sense to me :-). 

I am conscious of spending too much time investigating how best to structure the project rather than continuing work and using the existing structure. I believe our limited time is better spent progressing the current tasks. If you think that you would be able to research and provide your suggestion is less than 4 hours, please create a JIRA and present your proposed changes.

camilamacedo86:
It is fine 👍

PR #40, opened 2019-02-12
feat(swagger) : Generate swagger api from the source code (AEROGEAR-8518)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8518

## What
Implement changes to generated swagger api doc from the source code

## Why
Many reasons as, for example, the best practices and it better to keep maintained. 

## How
- It still using of https://github.com/go-swagger/go-swagger 
- By go-swagger generate the YAML file from the source code
- The YAML file now is created/updated when the `make setup` is made. 

### Check if the file will be generated automatically
1. Clone the project locally
2. Remove the swagger.YAML file of the project
3. Run the make setup command and check if the swagger.YAML file was created automatically

```shell
$ make setup
Installing errcheck dependence:
go get -u github.com/kisielk/errcheck
Setting up Git hooks:
ln -sf $PWD/.githooks/* $PWD/.git/hooks/
Installing application dependencies:
dep ensure
make build_swagger_api
Installing Swagger dep:
go get -u github.com/go-swagger/go-swagger/cmd/swagger
Updating Swagger api:
go generate ./cmd/mobile-security-service/main.go
```

### Check if the file is working on in the SwaggerUI
1. Get the row of the [swagger.YAML file](https://raw.githubusercontent.com/aerogear/mobile-security-service/8c8a300b599bbea26d61b7e6c07491ad42edc829/api/swagger.yaml) added here and check it in the [Swagger demo website](https://petstore.swagger.io/). 

screenshot 2019-02-12 at 18 47 31

2. Check if the file is valid in this UI/tool. 

screenshot 2019-02-12 at 18 48 04

## Checklist:

## Progress

- [X] Finished task

## Additional Notes

**NOTES** 

* The swagger definition adopted here is the same defined by @davidffrench in the previous file swagger.YAML and it is not what should be discussed in this PR since the goal here is just generated it from the source. 
* The coverage test is decreased because of the routers declaration. However, their implementation and tests are not part of this PR/JIRA/task.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ APP_NAME = mobile-security-service
 ORG_NAME = aerogear
 PKG = github.com/$(ORG_NAME)/$(APP_NAME)
 APP_FILE=./cmd/mobile-security-service/main.go
+APP_FILE_DIR=cmd/mobile-security-service
 TOP_SRC_DIRS = pkg
 PACKAGES     ?= $(shell sh -c "find $(TOP_SRC_DIRS) -name \\*_test.go \
                    -exec dirname {} \\; | sort | uniq")
@@ -22,6 +23,14 @@ LDFLAGS=-ldflags "-w -s -X main.Version=${TAG}"
 setup: setup_githooks
 	@echo Installing application dependencies:
 	dep ensure
+	make build_swagger_api
+
+.PHONY: build_swagger_api
+build_swagger_api:
+	@echo Installing Swagger dep:
+	go get -u github.com/go-swagger/go-swagger/cmd/swagger
+	@echo Updating Swagger api:
+	cd $(APP_FILE_DIR); swagger generate spec -m -o ../../api/swagger.yaml
 
 .PHONY: setup_githooks
 setup_githooks:
--- a/pkg/web/router/router.go
+++ b/pkg/web/router/router.go
@@ -4,9 +4,7 @@ import (
 	"github.com/aerogear/mobile-security-service/pkg/config"
 	"github.com/aerogear/mobile-security-service/pkg/web/apps"
 	"github.com/aerogear/mobile-security-service/pkg/web/middleware"
-
 	"github.com/labstack/echo"
-
 	"gopkg.in/go-playground/validator.v9"
 )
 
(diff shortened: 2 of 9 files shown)


pkg/web/router/router.go
```diff
@@ -28,5 +26,102 @@ func NewRouter(config config.Config) *echo.Echo {
...
+	// responses:
+	//   200:
+	//     description: successful operation
+	//     schema:
+	//       $ref: '#/definitions/App'
+	//   400:
+	//     description: Invalid id supplied
+	//   404:
+	//     description: App not found
+	r.GET("/apps/{id}", appsHandler.GetAppById) // TODO: Implement correctly the call of the method passing the parameters
```
davidffrench: @camilamacedo86 IS it possible for go-swagger to take more from the code. I always see the major benefit or auto-generated API docs, that we don't need to manually update information which is an additional step. With the comments here, we will need to update it if additional responses are added or other modifications.

Makefile
```diff
@@ -22,6 +22,14 @@ LDFLAGS=-ldflags "-w -s -X main.Version=${TAG}"
...
 	@echo Installing application dependencies:
 	dep ensure
+	make build_swagger_api
+
+.PHONY: build_swagger_api
+build_swagger_api:
+	@echo Installing Swagger dep:
+	go get -u github.com/go-swagger/go-swagger/cmd/swagger
+	@echo Updating Swagger api:
+	go generate $(APP_FILE)
```
davidffrench: This command will run all commands described by directives in our code. Your new swagger command is the only one in our code for now. If there is more in the future, this will end up running those as well.

pkg/web/router/router.go
camilamacedo86: @davidffrench in other languages as JAVA we can find the injections annotations to do it. The swagger will not parse the router implementation. It will parse comments and/or notations in all languages. 

However, in the case of the model, it is able to parse the struct as you can see [here](https://github.com/aerogear/mobile-security-service/pull/40/files#diff-d2e6ed7f6795bd3fdbcb783f1c3850a5R4). 

I liked the [beego](https://beego.me/) framework and how it works with swagger. It is using annotations which are more similar to java see [here](https://beego.me/blog/beego_api) how to it is done. However, for this project was adopted [echo](https://echo.labstack.com/) framework which is minimalist and has not this kind of feature. 

In this way, shows that the best approach is to use the [go-swagger](https://github.com/go-swagger/go-swagger) which is the same way that is done in the [aerogear-app-metrics](https://github.com/aerogear/aerogear-app-metrics/blob/master/pkg/web/router.go). 

Please, let me know if it still missing some clarification and/or you have any objection here. Also, please let me know if we can move forward and merge this pr.

Makefile
camilamacedo86: It is the same way that is done in the [aerogear-app-metrics](https://github.com/aerogear/aerogear-app-metrics/blob/master/cmd/metrics-api/metrics-api.go). Following some important info.

* The command should be called always in order to ensure that the swagger API will be updated/accurate. 
* In order to generate the swagger, API is required to execute it from the main.go directory. 

IMHO, for now, at least shows fine we use the `go generate` in order to achieve the expected result here. However, if you do not agree how would you like to see it implemented it? Have you any suggestion or could we move forward in this way then if be required we can change it in future?

pkg/web/router/router.go
davidffrench: Thanks for the clarification @camilamacedo86 . I am weighing up the pros and cons of including this.

*Pros*
- Pick up changes to a model on regeneration
- Swagger spec is easier to edit as it is now commented in our code

*Cons*
- Some models have fields which will never be returned in the response. This is confusing for someone just looking at the swagger api.
- The comments are verbose and still need to be updated manually for additional changes

Makefile
davidffrench: Is it possible to run the swagger command directly from the make file? `swagger generate spec -m -o ../../api/swagger.yaml`

pkg/web/router/router.go
camilamacedo86: Hi @davidffrench,

Thank you for your inputs.  Following some considerations regards the cons added here.

Regards 

> Some models have fields which will never be returned in the response. This is confusing for someone just looking at the swagger api

* The struct/objects mapped in the Swagger are just that ones which will be used in the endpoints and not all of them. 
* The metatag `swagger:model` should be added just in the structs that will be used by the API. 
* All struct that will be used as input/output of the RestAPI should be mapped and these structs don't necessary will be the database model. For example, an `/search` endpoint can return a list of items which will be exactly the database model or a list of items with a type that is constructed just for the response with just the relevant properties. (e.g ItemTO)

Regards

> The comments are verbose and still need to be updated manually for additional changes
* Besides I agree with the fact that it is verbose, it still lower than defined the YAML file manually as it was done before
* For routers the verbose are kind of similar, however, over the models, it is really better

Regards the pros added here.

> _Pros_
> 
> * Pick up changes to a model on regeneration
> * Swagger spec is easier to edit as it is now commented in our code   

I agree with them and I would like to add.

* It makes more intuitive the need of the update in the swagger syntax when a change is made
* It is verbose but less than defined the YAML file manually as it was
* The syntax is a common partner used and is at least intuitive besides for the router still verbose. 
* It is the commonly used approach was I could see in the projects with this stack/frameworks adopted
* It is done in the same way that we can find in the aerogear projects.

Makefile
camilamacedo86: It was changed as you requested. 👍 Thank you for the input.

pkg/web/router/router.go
camilamacedo86: Hi @davidffrench,

Thank you for your inputs.  Following some considerations and ideas.

**Regards _CONS_**

1. > Some models have fields which will never be returned in the response. This is confusing for someone just looking at the swagger api

* The struct/objects mapped for the Swagger API are just that ones which will be used in the endpoints and not all of them. 
* The metatag `swagger:model` should be added just in the structs that will be used by the API. 
* All struct that will be used as input/output of the RestAPI should be mapped and these structs don't necessary will be the database model. For example, an `/search` endpoint can return a list of items which will be exactly the database model struct or a list of items with a type that is constructed just for the input/response with the relevant properties only. (e.g ItemTO)

2. > The comments are verbose and still need to be updated manually for additional changes
* Besides I agree with the fact that it still verbose for the routers. It is less verbose and hard than define the full YAML file manually as it was done before

**Regards _PROS_**

> * Pick up changes to a model on regeneration
> * Swagger spec is easier to edit as it is now commented in our code   

I agree with them and I would like to add.

* It makes more intuitive the need of update the swagger syntax when a change is made
* It is verbose but less than defined the YAML file manually as it was
* The syntax is a commonly used and is at least intuitive besides for the router still verbose. 
* It is the commonly used approach that I could see in the projects with the same stack/frameworks adopted
* It is done in the same way that we can find in the aerogear projects.

So, wdyt? Could we move forward here or you would like to keep it manually done as it is?

davidffrench: Thank you and I agree, lets move forward with this.

PR #41, opened 2019-02-13
Add redux to manage UI state

damienomurchu:
## Motivation
Manage state better in UI (https://issues.jboss.org/browse/AEROGEAR-8527)

## What
Use redux store to store and manage UI app state

## Why
Manage state more effectively and efficiently

## How
Add redux lib, create single redux store, and add redux handlers and state to manage updates to app state

## Verification Steps

- Connect a component to the redux store & verify you are able to retrieve data from the store. For example, you can update `App.js` and check the console in your dev tools to verify the console.log prints row data:

```javascript
import React, { Component } from 'react';
import { connect } from 'react-redux';

class App extends Component {
  render() {
    console.log('rows', this.props.apps);
    return (
      <div className="App">
        <p>Mobile Security Service</p>
      </div>
    );
  }
}

const mapStateToProps = (state) => {
  return {
    apps: state.apps
  };
};

const mapDispatchToProps = (dispatch) => {
  return {};
};

export default connect(mapStateToProps, mapDispatchToProps)(App);
```

- Start application: `npm start`

- You should expect to see something like this in your browser console: 
```javascript
apps: [
    { id: 1000, app: 'App-A', versions: 3, clients: 1245, startups: 'male', birth_year: '1979', actions: null },
    { id: 1001, app: 'App-B', versions: 4, clients: 655, startups: 'male', birth_year: '1974', actions: null },
    { id: 1002, app: 'App-C', versions: 1, clients: 970, startups: 'female', birth_year: '1989', actions: null },
    { id: 1003, app: 'App-D', versions: 6, clients: 255, startups: 'male', birth_year: '1990', actions: null },
    { id: 1004, app: 'App-E', versions: 5, clients: 120, startups: 'female', birth_year: '1999', actions: null }
  ]
```
## Checklist:

## Progress

- [ ] Finished task
- [ ] TODO

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -3,9 +3,13 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "react": "^16.8.0",
-    "react-dom": "^16.8.0",
-    "react-scripts": "2.1.3"
+    "react": "16.8.0",
+    "react-dom": "16.8.0",
+    "react-redux": "6.0.0",
+    "react-scripts": "2.1.3",
+    "redux": "4.0.1",
+    "redux-logger": "3.0.6",
+    "redux-thunk": "2.3.0"
   },
   "scripts": {
     "start": "react-scripts start",
--- a/ui/src/configureStore.js
+++ b/ui/src/configureStore.js
@@ -1,0 +1,13 @@
+import { createStore, applyMiddleware } from 'redux';
+import thunkMiddleware from 'redux-thunk';
+import { createLogger } from 'redux-logger';
+import appsReducer from './reducers/apps';
+
+let middleware = [thunkMiddleware];
+if (process.env.NODE_ENV !== 'production') {
+  middleware = [...middleware, createLogger()];
+}
+
+export default function configureStore (preloadedState) {
+  return createStore(appsReducer, preloadedState, applyMiddleware(...middleware));
+}
--- a/ui/src/index.js
+++ b/ui/src/index.js
@@ -1,7 +1,26 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import { Provider } from 'react-redux';
+
 import './index.css';
 import App from './App';
+import configureStore from './configureStore';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+const preloadedState = {
+  apps: [
+    { id: 1000, app: 'App-A', versions: 3, clients: 1245, startups: 'male', birth_year: '1979', actions: null },
+    { id: 1001, app: 'App-B', versions: 4, clients: 655, startups: 'male', birth_year: '1974', actions: null },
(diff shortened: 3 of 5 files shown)


damienomurchu:
@laurafitzgerald just pinging you on PR here so you can see keep tabs on what I'm doing to implement redux. Using [MDC](https://github.com/aerogear/mobile-developer-console/blob/master/ui/src/reducers/apps.js) as a reference implementation, and figuring out how best to implement, thus the WIP status of PR for now :)

ui/package.json
```diff
@@ -5,7 +5,10 @@
   "dependencies": {
     "react": "^16.8.0",
     "react-dom": "^16.8.0",
-    "react-scripts": "2.1.3"
+    "react-scripts": "2.1.3",
+    "redux": "^4.0.1",
+    "redux-logger": "^3.0.6",
+    "redux-thunk": "^2.3.0"
   },
```
camilamacedo86: Is a good idea to use `^` here? Maybe, is not better it be `~`? Wdyt?

camilamacedo86: Sorry, I saw now that it is in WIP. The rest of the code shows great anyway.

damienomurchu: @camilamacedo86 np :) re versions these will be locked down before PR is merged

ui/src/index.js
```diff
@@ -1,7 +1,26 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import { Provider } from 'react-redux';
+
 import './index.css';
 import App from './App';
+import configureStore from './configureStore';
+
+const preloadedState = {
```
damienomurchu: This is just seed data and will be removed once we're fetching data from the db. Maybe it should live in it's own file though?

ui/src/reducers/apps.js
```diff
@@ -0,0 +1,6 @@
+export default function apps(state = [], action) {
+  switch (action.type) {
+    default:
```
damienomurchu: `default` action only action defined for now, as our app currently has no need for a reducer function yet

ui/src/configureStore.js
```diff
@@ -0,0 +1,13 @@
+import { createStore, applyMiddleware } from 'redux';
+import thunkMiddleware from 'redux-thunk';
```
damienomurchu: This is middleware to facilitate performing async requests with redux

damienomurchu:
@laurafitzgerald wow that was fast :D Are you happy to see this merged in now then?

laurafitzgerald:
@damienomurchu the only comment I have is that the preloaded store doesn't 100% match the state we'd expect to see in our application.

damienomurchu:
@laurafitzgerald yes, think we can evolve that as we evolve the UI. Ultimately the preLoadedState data will come from an api call, so I don't think we need to worry about the structure of the hardcoded seed data as it will go entirely soon.

Have you verified the PR? (Just want to close the JIRA & merge this as soon as you have :) )

PR #42, opened 2019-02-13
Updated /apps endpoint to get data from database (AEROGEAR-8526)

craicoverflow:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-8526

## What

Update the repository method for the [/apps](http://locahost:3000/api/apps) endpoint to get the data from the database.

## Why

We need to get persistent data from the database.

1. Make sure your database is populated with apps and associated versions.
2. Start the server.
3. Navigate to [/apps](http://locahost:3000/api/apps) and you should see those same apps in the database returned as JSON.
4. Set the `deleted_at` column on one of the apps in the database.
5. Reload the endpoint - does this app still show? If **yes** then verification has failed. 

## Checklist:

## Progress

- [x] Finished task

### Additional Notes

- The CircleCI build fails - this is because our integration test does not actually populate the database in order to properly test the response codes from the endpoints.


pkg/web/apps/apps_service.go
```diff
@@ -26,9 +27,31 @@ func NewService(repository Repository) Service {
 func (a *appsService) GetApps() (*[]models.App, error) {
 	apps, err := a.repository.GetApps()
 
+	// Check for errors and return the appropriate error to the handler
 	if err != nil {
```
davidffrench: All the error handling from line 30 to 41 can be simplified in my opinion by just returning the err from the repository layer and also checking if there are no apps in the repository layer and returning the `models.ErrNotFound` there

pkg/web/apps/apps_psql_repository.go
```diff
@@ -19,16 +22,59 @@ func NewPostgreSQLRepository(db *sql.DB) Repository {
...
-		ID:                    "a0874c82-2b7f-11e9-b210-d663bd873d93",
-		AppID:                 "com.aerogear.app1",
-		AppName:               "app1",
-		NumOfDeployedVersions: 1,
-		NumOfClients:          1,
-		NumOfAppLaunches:      1,
+	rows, err := a.db.Query(`SELECT id,app_id,app_name FROM app WHERE deleted_at IS NULL;`)
+
+	if err != nil {
+		log.Error(err)
```
camilamacedo86: @davidffrench it is a scenario that we could raise the error msg. For example, if the error msg here be "no connection" then this error msg could be thrown to the handler and returned in the response instead of the empty string. wdyt?

pkg/web/apps/apps_psql_repository.go
```diff
@@ -19,16 +22,59 @@ func NewPostgreSQLRepository(db *sql.DB) Repository {
...
+		if err != nil {
+			log.Error(err)
+		}
+	}()
+
+	apps := []models.App{}
+	for rows.Next() {
+		var a models.App
+		if err = rows.Scan(&a.ID, &a.AppID, &a.AppName); err != nil {
+			log.Error(err)
```
camilamacedo86: Should it not return an error?

pkg/web/apps/apps_psql_repository.go
```diff
@@ -19,16 +22,59 @@ func NewPostgreSQLRepository(db *sql.DB) Repository {
...
+		err := rows.Close()
+		if err != nil {
+			log.Error(err)
+		}
+	}()
+
+	versions := []models.Version{}
+
+	// iterate over the rows and add the data to the array of versions
+	for rows.Next() {
```
camilamacedo86: Has too many ways to solve the problems. In this imp you will spend too much memory to bring all data and to do the interactions, etc. It is possible to replace all in order to get the data required with just one select/query. You can do something like: 

```sql
SELECT
  tmp.app_pk as ID
  tmp.app_id as AppID,
  tmp.app_name as name,
  (count(*) group by tmp.app_pk,tmp.app_version_id) as NumOfDeployedVersions,
  (count(*) group by tmp.app_pk,tmp.device_id) as NumOfClients,
  (sum(tmp.launches) group by tmp.app_pk,tmp.app_version_id) as NumOfAppLaunches
FROM (
  SELECT
     a.id as app_pk,
     a.app_id as app_id,
     a.app_name as app_name,
     v.version_id as app_version_id,
     v.num_app_launches as launches,
     v.is_disabled as is_version_disabled,
     d.device_id as device_id,
     d.version_id as device_version_id
    FROM
      app as a,
      version as v,
      device as d
    WHERE
      AND a.deleted_at IS NOT NULL,
      AND a.app_id = v.app_id,
      AND v.version_id = d.version_id
) tmp
````
Also, if you find some issue doing as the above example then I'd like to suggest see how to use [INNER JOINS](https://www.postgresql.org/docs/8.3/tutorial-join.html).

File: pkg/web/apps/apps_psql_repository.go [hunk 5]
```diff
@@ -19,16 +22,59 @@ func NewPostgreSQLRepository(db *sql.DB) Repository {
...
+	defer func() {
+		err := rows.Close()
+		if err != nil {
+			log.Error(err)
+		}
+	}()
+
+	versions := []models.Version{}
+
+	// iterate over the rows and add the data to the array of versions
```
camilamacedo86: Hi @craicoverflow,

You don't need to do all these interactions. You can fetch all the data in just one query doing something as follows.  

```sql
  SELECT
     app.id as ID,
     app.app_id as AppID,
     app.app_name as AppName,
     (select count(DISTINCT v1.version_id) from version v1 where v1.app_id = app.app_id) as NumOfDeployedVersions,
     (select count(DISTINCT d1.device_id) from device d1 where d1.version_id in (
        SELECT DISTINCT v2.version_id FROM version v2 WHERE v2.app_id = app.app_id)
     ) as NumOfClients,
     (select sum(v3.num_app_launches) from version v3 where v3.app_id = app.app_id) as NumOfAppLaunches
    FROM
      app 
    WHERE
      app.deleted_at IS NOT NULL 
```
Also, you can do it with a better performance by using INNER JOINS. See [here](https://www.postgresql.org/docs/8.3/tutorial-join.html)

pkg/web/apps/apps_service.go
```diff
@@ -26,9 +27,31 @@ func NewService(repository Repository) Service {
...
+		}
+
+		return nil, models.ErrNotFound
+	}
+
+	if len(*apps) == 0 {
 		return nil, models.ErrNotFound
 	}
 
+	// iterate over the list of apps and get the deployed versions for each
```
camilamacedo86: All this part is not required if you fetch the data with one query as added in the above comment some example.

pkg/web/apps/apps_psql_repository.go
davidffrench: Where are you seeing the empty string? There is a current drawback of the generic models that there is no current way to input your own string.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -19,16 +22,59 @@ func NewPostgreSQLRepository(db *sql.DB) Repository {
...
+	if err != nil {
+		log.Error(err)
+		return nil, models.ErrInternalServerError
 	}
 
-	apps := []models.App{app1}
+	defer func() {
+		err := rows.Close()
+		if err != nil {
+			log.Error(err)
```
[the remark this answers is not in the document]
craicoverflow: This come after the SQL data has been scanned into the list of Apps. The error is because the connection to the database could not be closed, but we still got the apps and they will be returned at the end. Maybe this should be changed to a warning log?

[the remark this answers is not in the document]
craicoverflow: @davidffrench This come after the SQL data has been scanned into the list of Apps. The error is because the connection to the database could not be closed, but we still got the apps and they will be returned at the end. Maybe this should be changed to a warning log?

pkg/web/apps/apps_psql_repository.go
craicoverflow: @camilamacedo86 yes you are right, it should. Thanks :smile:

pkg/models/apps.go
```diff
@@ -3,11 +3,33 @@ package models
...
+}
+
+// SetNumOfAppLaunches calculates the total number of times
+// an app has been launches across each version
+func (a *App) SetNumOfAppLaunches() {
+	if a.DeployedVersions == nil {
+		return
+	}
+
+	for _, v := range *a.DeployedVersions {
```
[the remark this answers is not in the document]
austincunningham: SQL can do it but don't think the sql module in go can handle it

pkg/web/apps/apps_psql_repository.go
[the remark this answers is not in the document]
davidffrench: Thanks for the clarification. I think it should stay as an Error in this case.

pkg/web/apps/apps_psql_repository.go
camilamacedo86: Wdyt over create a task for us to improve the generic model's error class and allow it? IHMO it is a high import for a dev get the error message to know what is going wrong and be able to solve it.

pkg/web/apps/apps_psql_repository.go
craicoverflow: Hi @camilamacedo86 you are right. I have updated this now. One thing though - INNER JOINS do not work for this, if a row in the app table does not have any associated versions then the query does not return that app at all. For this reason we must use LEFT join.

camilamacedo86: Hi @craicoverflow,

Yes. You have LEFT JOIN and OUT JOIN and INNER JOIN. Regards, this scenario I believe that you are correct and it will need the LEFT JOIN. 

Note that my comment was to guide you to look for how you can solve it in the database instead of doing this interaction then since you reply it hows that the goal was achieved. :-)

pkg/web/apps/apps_psql_repository.go
craicoverflow: @camilamacedo86 I agree this should be done at some point. What we have now is limited. [Here is](https://medium.com/@sebdah/go-best-practices-error-handling-2d15e1f0c5ee) a good implementation I came across yesterday.

pkg/models/apps.go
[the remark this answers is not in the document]
craicoverflow: @austincunningham this can be done. I can get calculated columns fine but cannot map a related table to Go in a single query.

pkg/web/apps/apps_psql_repository.go
camilamacedo86: @craicoverflow good post. I will create the JIRA and add it there so. tks for the support

camilamacedo86: https://issues.jboss.org/browse/AEROGEAR-8542

pkg/web/apps/apps_psql_repository.go
```diff
@@ -19,16 +22,75 @@ func NewPostgreSQLRepository(db *sql.DB) Repository {
...
-		AppID:                 "com.aerogear.app1",
-		AppName:               "app1",
-		NumOfDeployedVersions: 1,
-		NumOfClients:          1,
-		NumOfAppLaunches:      1,
+	rows, err := a.db.Query(`
+	SELECT a.id,a.app_id,a.app_name,
+	COUNT(v.id) as num_of_deployed_versions, 
+	COALESCE(SUM(v.num_of_app_launches),0) as num_of_app_launches,
+	COALESCE(SUM(v.num_of_clients),0) as num_of_clients
```
camilamacedo86: Note that here the number of clients need group by the version_id.
One device/mobile can change the version installed of the same app.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -19,16 +22,75 @@ func NewPostgreSQLRepository(db *sql.DB) Repository {
...
 
-	apps := []models.App{app1}
+	if len(apps) == 0 {
+		return nil, models.ErrNotFound
+	}
 
 	return &apps, nil
 }
+
+func (a *appsPostgreSQLRepository) GetAppVersionsByAppID(id string) (*[]models.Version, error) {
```
camilamacedo86: I do not understand why you add this func here. You can create the struct Apps with just one query. In this same way, you can create just one query to return the struct app for some app_id. Also, I don't think it is part of this task anyway.

pkg/web/apps/apps_psql_repository_test.go
```diff
@@ -1,56 +1,241 @@
...
+
+	timestamp := "2019-02-15T09:38:33+00:00"
+
+	// Insert an app where the deleted_at column is set
+	sqlmock.NewRows(cols).AddRow(mockApps[0].ID, mockApps[0].AppID, mockApps[0].AppName, timestamp)
+
+	// Insert 2 apps which are not soft deleted
+	rows := sqlmock.NewRows(cols).AddRow(mockApps[1].ID, mockApps[1].AppID, mockApps[1].AppName, "").AddRow(mockApps[2].ID, mockApps[2].AppID, mockApps[2].AppName, "")
+
+	query := `
```
camilamacedo86: Why have 2 places with the same implementation/duplicate code? It shows that should be centralized.

pkg/web/apps/apps_psql_repository_test.go
```diff
@@ -1,56 +1,241 @@
...
+			AppName: "Mobile App Three",
+		},
+	}
+
+	timestamp := "2019-02-15T09:38:33+00:00"
+
+	// Insert 3 apps which are soft deleted
+	sqlmock.NewRows([]string{"id", "app_id", "app_name", "deleted_at"}).AddRow(mockApps[0].ID, mockApps[0].AppID, mockApps[0].AppName, timestamp).AddRow(mockApps[1].ID, mockApps[1].AppID, mockApps[1].AppName, timestamp).AddRow(mockApps[2].ID, mockApps[2].AppID, mockApps[2].AppName, timestamp)
+
+	query := `
```
camilamacedo86: The same over the above comment. Many places with the same code/query.

pkg/web/apps/apps_repository.go
```diff
@@ -7,4 +7,5 @@ import (
 // Repository represent the app's repository contract
 type Repository interface {
 	GetApps() (*[]models.App, error)
+	GetAppVersionsByAppID(id string) (*[]models.Version, error)
```
camilamacedo86: Should not be here return GetAppsByID() which indeed will be part of /apps/{id} router and task and not this one?

pkg/web/apps/apps_psql_repository.go
craicoverflow: @camilamacedo86 it is possible to retrieve all apps and versions but it does seem possible to map the rows from the `version` table into Go within this same query. It seems to be a limitation of the standard SQL Go library.

If you can find an example where this is done I would be happy to implement it, but have tried and could not find any.

Also, this is part of this task as this data is required for the `/apps` endpoint.

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: This is required for [go-sqlmock](https://github.com/DATA-DOG/go-sqlmock). It compare this SQL query to the one in the actual repository function and fails if they are mismatched.

[Example](https://github.com/bxcodec/go-clean-arch/blob/master/article/repository/mysqlarticle_test.go#L59)

pkg/web/apps/apps_repository.go
craicoverflow: @camilamacedo86 `GetAppVersionsByAppID` is required for the `/apps` endpoint. @austincunningham is working on the `GetAppsByID` endpoint

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: See above

pkg/web/apps/apps_psql_repository_test.go
camilamacedo86: The point here is that we should centralize the code in one place. In the way that it is implemented if be required change the query then need to look for many places since the same query/string/code can be found in more than 2 places. Is it make sense?

pkg/web/apps/apps_repository.go
camilamacedo86: Hi @craicoverflow,

What is the struct that will return in the endpoint? Then, you don't need to make queries to return this structure no matter what tables each will be looking for. You just need one query which will return all data required which will be used to create the ResponseObject. 

Important: I understand that you need to return in the endpoint apps/ is the app struct.

camilamacedo86:
@craicoverflow I think it will be easier if you close the comments over what was already changed/solved in this PR.

pkg/web/apps/apps_psql_repository_test.go
```diff
@@ -1,56 +1,241 @@
...
-	"github.com/aerogear/mobile-security-service/pkg/db"
-
 	"github.com/aerogear/mobile-security-service/pkg/models"
+	sqlmock "gopkg.in/DATA-DOG/go-sqlmock.v1"
 )
 
-func Test_appsPostgreSQLRepository_GetApps(t *testing.T) {
-	c := config.Get()
-	dbConn, err := db.Connect(c.DB.ConnectionString, c.DB.MaxConnections)
+func Test_appsPostgreSQLRepository_GetAppsWillReturnTwoApps(t *testing.T) {
```
camilamacedo86: The mock data or func shows that is not working. See the CircleCI issue. 

screenshot 2019-02-15 at 17 12 11

pkg/web/apps/apps_psql_repository.go
camilamacedo86: Please, look my comment [here](https://github.com/aerogear/mobile-security-service/pull/42#discussion_r257296253)

pkg/web/apps/apps_psql_repository.go
camilamacedo86: closing as resolved.

camilamacedo86:
@craicoverflow you have been doing amazing work here.
I think that more some changes and it will be perfect to be merged.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -19,16 +22,59 @@ func NewPostgreSQLRepository(db *sql.DB) Repository {
...
+			log.Error(err)
+		}
+
+		apps = append(apps, a)
+	}
 
 	return &apps, nil
 }
+
+func (a *appsPostgreSQLRepository) GetAppVersions(id string) (*[]models.Version, error) {
```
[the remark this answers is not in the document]
craicoverflow: @davidffrench done.

pkg/web/apps/apps_psql_repository_test.go
camilamacedo86: Is it(Test_appsPostgreSQLRepository_GetAppsWillReturnTwoApps) a really good name? 
Shows that we are not following here thee go convention. See [here](https://golang.org/pkg/testing/#pkg-overview). I understand that it could be something as TestGetAllAppsFromRespository. Wdyt?

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: @camilamacedo86 I partially agree with you - there is a duplication of content here. However this would make our unit tests dependent on an external state which is not good practice either. I am not sure of the best approach here.

Also, notice that the query string is different in the test file. This is because this is is a regex and not a pure SQL query string. It does a compare to the actual query being performed.

I've taken a step to reduce duplication by having the regex strings to the top of the test file.

camilamacedo86: HI @craicoverflow,

Following my IHMO. If the same code is duplicated many times in this file only. Then a simple solution can add a global const var with this info that can be used just in this class/routine.

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: This is because the database which is created during integration tests does not have any seed data. This will need to be added. This is not directly caused by the mock data - it is separate but it will need to be added.

craicoverflow: The VS Code extension generates unit test names like this. I appended the function name with `ShouldReturnNoApps` because it describes what the function is expected to do. I can't from the link you provided where it says above is bad?

> Multiple example functions for a package/type/function/method may be provided by appending a distinct suffix to the name. The suffix must start with a lower-case letter.

pkg/web/apps/apps_repository.go
craicoverflow: Hi @camilamacedo86 ,

I believe it not possible with the standard SQL library in Go to map the apps table and the joined table (version) into a Go struct in a single query. It is possible to retrieve the data with SQL, the limitation is in **mapping it to Go**.

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: Hi @camilamacedo86 this is what I suggested above:

> I've taken a step to reduce duplication by having the regex strings to the top of the test file.

pkg/web/apps/apps_psql_repository_test.go
camilamacedo86: It was before `Test_appsPostgreSQLRepository_GetAppsWillReturnTwoApps` and not it is `Test_appsPostgreSQLRepository_GetApps` which indeed shows better. 👍 You can consider it solved.

pkg/web/apps/apps_psql_repository.go
craicoverflow: I don't fully see why it needs to group by `device_id` and `version_id`? The `version ` table has the number of app launches per version. We can just sum all of these values per app. What is the need to do an extra join to the `device` table?

pkg/web/apps/apps_psql_repository_test.go
camilamacedo86: Closing it as resolved so.

pkg/web/apps/apps_repository.go
camilamacedo86: Hi @craicoverflow,

Sorry, but I do not understand what is the limitation here. Should something as the following example not work here?

```go
func TestQuery(t *testing.T) {
    /* . . . */
    rows, err := db.Query("SELECT 
                a.id ID,
               a.app_id AppID,
               a.app_name AppName,
	      COUNT(v.id) as num_of_deployed_versions NumOfDeployedVersions, 
	      COALESCE(SUM(v.num_of_app_launches),0) as num_of_app_launches NumOfClients,
	       COALESCE(SUM(v.num_of_clients),0) as num_of_clients NumOfAppLaunches
	FROM app as a 
	LEFT JOIN version as v on a.app_id = v.app_id 
	WHERE a.deleted_at IS NULL 
	GROUP BY a.id;`)")
    if err != nil {
            t.Fatalf("Query: %v", err)
    }

    type row struct {
            id  int
            AppID string
           // TODO the rest of impl.  
    }   

    got := []row{}
    for app.Next() {
            var r row
            err = rows.Scan(&r.id, &r.AppID, &r.AppName, &r.NumOfDeployedVersions, &r.NumOfClients, &r.NumOfAppLaunches )
            if err != nil {
                    t.Fatalf("Scan: %v", err)
            }
            got = append(got, r)
    }
    /* . . . */
} 
```

pkg/web/apps/apps_psql_repository.go
camilamacedo86: Because the number of clients is the number of mobile devices with the application installed. If I have the app installed than in the database will have the version + my device_id + app_id. However, if I update my device/mobile for a newer version of this app, then in the database will have another row as newer_version + device_id + app_id. In this way, if you do not group by here it will return 2 when in the reality the app is installed just in one mobile device and not in 2. Is it make sense? 

PS>:I understand that the apps/ will return the total per app and not per version.

pkg/web/apps/apps_psql_repository_test.go
camilamacedo86: HI @craicoverflow,

Just to complement and share. If we are using the same query in more than one routine we could centralize it doing something like the following example.



```
type QueryStore struct {
    db          *sql.DB
    prepared    map[string]*sql.Stmt   // you might want to use an Enum type instead of string for added safety
}

func NewQueryStore(db *sql.DB) *QueryStore {
    q := &QueryStore{db: db, prepared: map[string]*sql.Stmt{}}
    
    // add your prepared statements to the map
    stmt, err := db.Prepare("bar")
    checkErr(err)
    q.prepared[foo] = stmt

    // etc

    return q
}

func (s QueryStore) getNames(param string) []string{
    rows, err := s.prepared[foo].Query(param)
    // error checking
    var results []string
    // row scanning
    // more error checking
    // etc...
    return results
} 
```

pkg/web/apps/apps_repository.go
craicoverflow:
```go
func (a *appsPostgreSQLRepository) GetApps() (*[]models.App, error) {
	rows, err := a.db.Query(`
	SELECT a.id,a.app_id,a.app_name,
	COALESCE(COUNT(DISTINCT v.id),0) as num_of_deployed_versions,
	COALESCE(SUM(DISTINCT v.num_of_app_launches),0) as num_of_app_launches,
	COALESCE(SUM(DISTINCT v.num_of_clients),0) as num_of_clients
	FROM app as a LEFT JOIN version as v on a.app_id = v.app_id 
	WHERE a.deleted_at IS NULL 
	GROUP BY a.id;`)

	if err != nil {
		log.Error(err)
		return nil, models.ErrInternalServerError
	}

	defer func() {
		err := rows.Close()
		if err != nil {
			log.Error(err)
		}
	}()

	apps := []models.App{}
	for rows.Next() {
		var a models.App
		if err = rows.Scan(&a.ID, &a.AppID, &a.AppName, &a.NumOfDeployedVersions, &a.NumOfAppLaunches, &a.NumOfClients); err != nil {
			log.Error(err)
		}

		apps = append(apps, a)
	}

	if len(apps) == 0 {
		return nil, models.ErrNotFound
	}

	return &apps, nil
}
```

Apart from the fact that your above example is in a test function, what is different?

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: This may be a tad out of scope of this task as it requires a whole new line of investigation.

pkg/web/apps/apps_psql_repository.go
craicoverflow: @camilamacedo86 I had a conversation with @davidffrench as he has more insight into this. The purpose of `version.num_of_clients` is to keep track of the number of clients for that specific version. So as it stands the query is retrieving the correct information.

pkg/web/apps/apps_repository.go
camilamacedo86: So, what is not possible?

craicoverflow: The original comment in this thread is regarding `GetAppVersionsByAppID` so I believed you were talking about something different. We are no longer talking about `GetAppVersionsByAppID`.

pkg/web/apps/apps_service.go
craicoverflow: @davidffrench done.

pkg/models/versions.go
```diff
@@ -2,12 +2,17 @@ package models
...
-	AppID            string `json:"appId"`
-	Disabled         bool   `json:"disabled"`
-	DisabledMessage  string `json:"disabledMessage,omitempty"`
-	NumOfClients     int64  `json:"numOfClients"`
-	NumOfAppStartups int64  `json:"numOfAppStartups"`
+	ID               int64          `json:"id"`
+	Version          string         `json:"version"`
+	AppID            string         `json:"appId"`
+	Disabled         bool           `json:"disabled"`
+	DisabledMessage  sql.NullString `json:"disabledMessage,omitempty"`
```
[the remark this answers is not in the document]
craicoverflow: I've made a change for this. It's not ideal but it is better than returning an object to the client for any nullable database value.

https://github.com/aerogear/mobile-security-service/pull/42/commits/3ef3fe94efb2ad5e993e363f56d8512263db22b8

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: @camilamacedo86 I've added a function to seed the database before running the integration tests. This should succeed now.

pkg/models/app.go
```diff
@@ -3,11 +3,11 @@ package models
...
-	NumOfDeployedVersions int64     `json:"numOfDeployedVersions,omitempty"`
-	NumOfClients          int64     `json:"numOfClients,omitempty"`
-	NumOfAppLaunches      int64     `json:"numOfAppLaunches,omitempty"`
-	DeployedVersions      []Version `json:"deployedVersions,omitempty"`
+	ID                    string     `json:"id,omitempty"`
+	AppID                 string     `json:"appId,omitempty"`
+	AppName               string     `json:"appName,omitempty"`
+	NumOfDeployedVersions int        `json:"numOfDeployedVersions"`
+	NumOfClients          int        `json:"numOfClients"`
+	NumOfAppLaunches      int64      `json:"numOfAppLaunches"`
```
camilamacedo86: Why just this one is the type int64 as the NumOfDeployedVersions and NumOfClients are int?

craicoverflow: Good spot, updated.

pkg/web/apps/apps_repository.go
camilamacedo86: I am was. I do not think it should be here in this PR since its not part of this task. However, it is fine. I will close this comment because shows that it lost the focus with so misunderstands.

craicoverflow: The data is required for the `/apps` endpoint so it is a requirement for this PR.

pkg/web/apps/apps_psql_repository_test.go
```diff
@@ -1,56 +1,234 @@
...
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("appsPostgreSQLRepository.GetApps() = %v, want %v", got, tt.want)
-			}
-		})
+		t.Fatalf("Unexpected error opening a stub database connection: %v", err)
+	}
+
+	defer db.Close()
+
+	mockApps := []models.App{
```
camilamacedo86: Is not the same mock twice in the code? If yes, wdyt over centralize it? Wdyt over add a global var with this value? in fact, shows that it could be in fixtures.go. Wdyt?

pkg/models/app.go
```diff
@@ -3,11 +3,11 @@ package models
...
-	AppID                 string    `json:"appId"`
-	AppName               string    `json:"appName"`
-	NumOfDeployedVersions int64     `json:"numOfDeployedVersions,omitempty"`
-	NumOfClients          int64     `json:"numOfClients,omitempty"`
-	NumOfAppLaunches      int64     `json:"numOfAppLaunches,omitempty"`
-	DeployedVersions      []Version `json:"deployedVersions,omitempty"`
+	ID                    string     `json:"id,omitempty"`
+	AppID                 string     `json:"appId,omitempty"`
+	AppName               string     `json:"appName,omitempty"`
+	NumOfDeployedVersions int        `json:"numOfDeployedVersions"`
```
davidffrench: @craicoverflow Hwo come the `omitempty` has been changed from here to the ID, AppID and AppName fields?

pkg/web/apps/apps_psql_repository_test.go
```diff
@@ -1,56 +1,234 @@
...
+
+func Test_appsPostgreSQLRepository_GetAppVersionsByAppID(t *testing.T) {
+	db, mock, err := sqlmock.New()
+	if err != nil {
+		t.Fatalf("Unexpected error opening a stub database connection: %v", err)
+	}
+
+	defer db.Close()
+
+	mockVersions := []models.Version{
```
camilamacedo86: I know that you folks decided it and probably you will not change. However, I would like to say that I do not agree besides know that you will move forward in this way. The versions should not be the object of the app. They are different things. 
IHMO:
* The first page will use the endpoint /apps which should return a list of the object app. 
* then, is if someone clicks on in the app of the list the versions of that app with the totals should appear. So, it can't be the same object used in the list, it should be another object, for example, AppDetails with the attributes that are used in the details of an app.

davidffrench: I do not understand your point @camilamacedo86 . The landing UI screen is a high-level list of all apps(`/apps`). The detailed UI screen is in the context of a single app (`/apps/{id}`) and all information including sub-information should be displayed in this screen.

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: @camilamacedo86 yes this makes sense. I will do this now.

pkg/web/apps/apps_psql_repository_test.go
craicoverflow: @camilamacedo86 nobody will see your message except me, you should CC the relevant people here or say this in Google Hangouts Chat so people are aware of your concerns.

pkg/models/app.go
craicoverflow: @davidffrench Hmm I am not sure why it got added to the ID, AppID and AppName fields - possibly because this PR has merged a lot of changes from other branches in the past week or so and it got overwritten. They should not have `omitempty` though so I will remove.

davidffrench: Thanks @craicoverflow . What was the reason for removing them from `NumOfDeployedVersions`, `NumOfClients` and `NumOfAppLaunches`?

craicoverflow: Because if their values are 0 then the values will not be sent back with in the payload. We want those values even if they will be displayed on the client UI.

davidffrench: This may be a problem as we did not want them sent back on the `/apps/{id}` response which will be using the same model. It is not a big deal though if they are sent back in that request with no values.

craicoverflow: Pointers fixes this problem, updated PR.

https://stackoverflow.com/a/38487668

davidffrench: That is really good to know, thank you

davidffrench:
@craicoverflow After verifying now.

Verified 204 returned with no content in the database and with all content in the database with `deletedAt` set.

Verified content returning and values calculating correctly.

One small thing I noticed. The deployedVersions should not be returned in this call.

craicoverflow:
> @craicoverflow After verifying now.
> 
> Verified 204 returned with no content in the database and with all content in the database with `deletedAt` set.
> 
> Verified content returning and values calculating correctly.
> 
> One small thing I noticed. The deployedVersions should not be returned in this call.

@davidffrench this has been updated. Can you test again when you get a chance?

craicoverflow:
> > @craicoverflow After verifying now.
> > Verified 204 returned with no content in the database and with all content in the database with `deletedAt` set.
> > Verified content returning and values calculating correctly.
> > One small thing I noticed. The deployedVersions should not be returned in this call.
> 
> @davidffrench this has been updated. Can you test again when you get a chance?

@davidffrench apologies I misunderstood. I have now removed the deployed versions completely from the `/apps` endpoint.

PR #46, opened 2019-02-15
fix(swagger) - Not allow merges with the swagger documentation outdated

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8541

## What
* Not allow merges with the swagger documentation outdated

## Why
* the file in the repo can be used to check it documentation
* for docker image be build with the correct information
* to not allow devs check it outdated when the steps described in the readme to check the swagger api doc locally be executed

## How
* Create a script to verify it and called in the CI
* Commit the changes made in the swagger doc api

## Verification Steps
* Check if the CI finish with success here 
* Run the script `./scripts/check-if-swagger-api-file-is-update.sh` locally if the `api/swagger.YAML` is not outdated then the script cannot return an error. Otherwise, it needs to return the error. Following an example of how to do make it fails. 
1. Add more one declaration into the apps struct 
2. Run the command `./scripts/check-if-swagger-api-file-is-update.sh` 
3. Following the expected result.

```shell
Checking swagger API documentation ...
Installing Swagger dep:
go get -u github.com/go-swagger/go-swagger/cmd/swagger
Updating Swagger API:
cd cmd/mobile-security-service; swagger generate spec -m -o ../../api/swagger.yaml
The REST API Swagger doc was updated and it was not committed.
Please, run 'make setup' or 'build_swagger_api' and commit the changes in the file 'api/swagger.yaml'.
```
## Checklist:

- [x] Code has been tested locally by PR requester
- [ ] Changes have been successfully verified by another team member 

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes
 it was also tested with the CI.

screenshot 2019-02-15 at 11 47 22

PR #49, opened 2019-02-18
fix(CI) : CI test which is calling the script code-check-install-dep.sh do not fail when the script fails (AEROGEAR-8550)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8550

## What
CI test which is calling the script code-check-install-dep.sh do not fail when the script fails

## Why
To not allow merge PRs which are not passing in the defined rules.

## How
The termination `set -o errexit` is not returning exit as 1 error to CircleCi. The solution here is just to replace it for `exit 1` which will works

### How to check the issue? 
1. Locally before this PR/in the current project add a file and try to commit it with a message which will not pass in the script. 
2. Check the error and now commit it with ` --no-verify`
3. Push it to a PR request in the repo and check that all CI tests will be finished with success.

### How to check that `exit 1`?  
* You can see that exit 1 worked well as expected for other checks. See [here](https://github.com/aerogear/mobile-security-service/pull/46)
* You can try to do a Pull Request with this fix + a commit with a message without the format which will cause this check fail as described above. 

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/scripts/commit-filter-check.sh
+++ b/scripts/commit-filter-check.sh
@@ -59,7 +59,7 @@ commit_message_check (){
             echo $gitmessage
             echo " "
             rm shafile.txt >/dev/null 2>&1
-            set -o errexit
+            exit 1
       else
             echo "$messagecheck"
             echo "'$i' commit message passed"


scripts/commit-filter-check.sh
```diff
@@ -59,7 +59,7 @@ commit_message_check (){
             echo $gitmessage
             echo " "
             rm shafile.txt >/dev/null 2>&1
-            set -o errexit
+            exit 1
```
austincunningham: Doesn't match the circle ci docs https://circleci.com/docs/2.0/using-shell-scripts/#set-error-flags

austincunningham: Have no issue with the change, will try and verify

camilamacedo86: Thank you 👍

camilamacedo86:
@craicoverflow really thank you for your time and review here.

PR #50, opened 2019-02-18
feat(swagger): Added a Docker Compose entry for Swagger (AEROGEAR-8548)

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8548

## What

Added Swagger container to Docker compose file.

## Why

To make it easier to run and view the API documentation.

## Verification Steps
 
1. Run `docker-compose up swagger`
2. Go to localhost:8080
3. Is Swagger running at this address?

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,3 +20,9 @@ services:
       - "3000:3000"
     depends_on:
       - db
+  swagger:
+    image: swaggerapi/swagger-ui:v3.0.5
+    ports:
+      - 8080:8080
+    environment:
+      API_URL: https://raw.githubusercontent.com/aerogear/mobile-security-service/master/api/swagger.yaml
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -19,4 +19,10 @@ services:
...
     depends_on:
-      - db
\ No newline at end of file
+      - db
+  swagger:
+    image: swaggerapi/swagger-ui:v3.0.5
+    ports:
+      - 8080:8080
+    environment:
+      API_URL: https://raw.githubusercontent.com/aerogear/mobile-security-service/master/api/swagger.yaml
```
camilamacedo86: Could it be changed to get the data from the local file in the dev env? In this way, the UI will reflect the changes performed locally.

craicoverflow: I don't believe so. This came up before and we came to the conclusion that it could not be done.

davidffrench: Swagger needs the file to be served over http from what I found.

camilamacedo86: However, is not possible to retrieve the file by something as localhost:3000/api ...?

davidffrench: If there is a web server running locally, that is indeed possible.

PR #52, opened 2019-02-18
Add implementation and change into the readme in order to define how to create the mocks for the interfaces (AEROGEAR-8530)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8530

## What
* Add implementation and change into the readme in order to define how to create the mocks for the interfaces

## Why
* The unit tests need to mock the data and not call the database

## How
- Using the lib [moq](https://github.com/matryer/moq) as it was defined
- Updating the readme to make clear how to create the mocks

## Verification Steps
- Check if after the mock it still passing in the CI. ( unit tests )

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

I understand that we need to improve the unit tests, however, it should be part of another task. The idea here was just add in the project a way to add the generate and use the mocks with this lib/dep.  

Also, shows that the only unit test added so far in the project, which is not in the repository layer, was updated in order to use the mock. The rest of the tests implemented so far shows to be integrations tests.


Makefile
```diff
@@ -66,12 +66,18 @@ test-all: test-unit
 .PHONY: test
 test: test-unit
 
+.PHONY: mock_setup
+mock_setup:
+	@echo Installing mock dependence:
+	go get -u github.com/matryer/moq
```
davidffrench: What is the best practice of Go dev dependencies? I am not a big fan of adding another make command to install one dependency.

pkg/web/apps/app_service_moq_test.go
```diff
@@ -0,0 +1,70 @@
+// Code generated by moq; DO NOT EDIT.
```
davidffrench: What benefit are we getting out of this additional file?

Makefile
camilamacedo86: We can remove this from the make file and have it just in the readme as an explanation.

davidffrench: I am not sure how dev dependencies should be installed in Go. In Node.js it splits it logically in the `package.json`. Do you know how this should be done in Go?

pkg/web/apps/app_service_moq_test.go
camilamacedo86: The file has the mock implementation. it is called/used in the tests. You can check if we remove this file then the test will fail because no longer the implementation is in the project. 

```
$ make test
Running tests:
GOCACHE=off go test -cover \
	  github.com/aerogear/mobile-security-service/pkg/config github.com/aerogear/mobile-security-service/pkg/db github.com/aerogear/mobile-security-service/pkg/httperrors github.com/aerogear/mobile-security-service/pkg/test github.com/aerogear/mobile-security-service/pkg/web/apps
ok  	github.com/aerogear/mobile-security-service/pkg/config	0.043s	coverage: 100.0% of statements
?   	github.com/aerogear/mobile-security-service/pkg/db	[no test files]
# github.com/aerogear/mobile-security-service/pkg/web/apps [github.com/aerogear/mobile-security-service/pkg/web/apps.test]
pkg/web/apps/apps_service_test.go:22:23: undefined: AppServiceMock
ok  	github.com/aerogear/mobile-security-service/pkg/httperrors	0.056s	coverage: 70.4% of statements
?   	github.com/aerogear/mobile-security-service/pkg/test	[no test files]
FAIL	github.com/aerogear/mobile-security-service/pkg/web/apps [build failed]
make: *** [test-unit] Error 2
```

screenshot 2019-02-19 at 12 01 46

Is it make sense?

Makefile
camilamacedo86: Done 👍

pkg/web/apps/app_service_moq_test.go
davidffrench: Yes it makes sense. I wonder if it can be all combined into the same test file. I will be tackling this for my own endpoint in the next day or so and will learn more about it.

pkg/web/apps/apps_http_handler.go
```diff
@@ -10,24 +10,24 @@ import (
 )
 
 type (
-	// HTTPHandler instance
-	HTTPHandler struct {
-		Service AppService
+	// AppsHTTPHandler instance
+	AppsHTTPHandler struct {
```
craicoverflow: Why did you change this name? It goes against the Go recommendation.

> A package name and its contents' names are coupled, since client code uses them together. When designing a package, take the client's point of view.

> Avoid stutter. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the `http` package is called Server, not `HTTPServer`. Client code refers to this type as `http.Server`, so there is no ambiguity.

[The Go Blog - Package names](https://blog.golang.org/package-names)

README.adoc
```diff
@@ -177,6 +177,36 @@ The `Makefile` provides commands for building and testing the code. Some depende
...
+$ go get -u github.com/matryer/moq
+----
+
+=== Creating mocks
+
+Execute the command the following command in the directory where is the interface.
+
+[source,shell]
+----
+$ moq -out <name_of_mock_file> . <name of interface>
```
craicoverflow: If we want to run this from the project root, this should show us how. Might I suggest a small change:

```sh
$ moq -out <name_of_mock_file> ./path/to/package <name of interface>
```

pkg/httperrors/httperrors.go
```diff
@@ -39,7 +39,7 @@ var codes = map[int]string{
 	500: "Internal Server Error",
 	501: "Not Implemented",
 	502: "Bad Gateway",
-	503: "Service Unavailable",
+	503: "AppService Unavailable",
```
craicoverflow: This should be a generic default message. This affects areas of the server. It should remain as the default and if we want to at some point return "AppService Unavailable" it can be supplied as an argument.

https://www.lifewire.com/503-service-unavailable-explained-2622940

craicoverflow: This should be a generic default message. This affects areas of the server. It should remain as the default and if we want to at some point return "AppService Unavailable" it can be supplied as an argument.

https://www.lifewire.com/503-service-unavailable-explained-2622940

pkg/web/apps/app_service_moq_test.go
craicoverflow: What single file would they be in @davidffrench?

pkg/web/apps/apps_http_handler.go
camilamacedo86: makes sense @craicoverflow tks

README.adoc
camilamacedo86: The command needs to be executed where the Interface is. I am afraid that will not work.

craicoverflow: The following command worked for me:

```sh
$ moq -out ./pkg/web/apps/mocks_test.go ./pkg/web/apps AppService
```

craicoverflow:
I am wondering do we also need to mock the handlers in the same way?

@camilamacedo86 @davidffrench

pkg/web/apps/app_service_moq_test.go
davidffrench: I would rather if it was possible to have all unit tests for a single file contained in a single unit test file. This does not seem to be how the mocks work, I need to look into them more.

craicoverflow: The mock file doesn't have any unit tests though - it is different. It mocks the service so that the tests can interact with them as if they were the real service. It would be misleading having it in the test file.

See how Integreatly use it: [realm/handler_moq.go](https://github.com/integr8ly/keycloak-operator/blob/bb03fa20279a3f88159f469f308c0d1da8c62a05/pkg/keycloak/realm/handler_moq.go)

davidffrench: Thank you @craicoverflow

davidffrench:
> I am wondering do we also need to mock the handlers in the same way?
> 
> @camilamacedo86 @davidffrench

Yes, in my opinion.

pkg/web/apps/app_service_moq_test.go
camilamacedo86: @davidffrench if we combine the file into the test then it will lose the advantage to use the lib/dep. The lib dep will create/update the mock file automatically to be used in the tests. Could we agree in move forward in this way, at least for now?

Makefile
```diff
@@ -72,6 +72,7 @@ test-unit:
 	GOCACHE=off go test -cover \
 	  $(addprefix $(PKG)/,$(PACKAGES))
 
+
```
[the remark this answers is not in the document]
camilamacedo86: 👍

pkg/web/apps/app_service_moq_test.go
davidffrench: Absolutely. Please proceed

camilamacedo86:
@davidffrench and @craicoverflow,

The dep just mock Interfaces. The handler currently is not implemented as an interface. See [here](https://github.com/aerogear/mobile-security-service/blob/master/pkg/web/apps/apps_http_handler.go). I agree that we also should do it for the handler. However, first, we need to improve/change the handler implementation/ Then we can mock it too.

pkg/web/apps/app_service_moq_test.go
camilamacedo86: Thank you 👍

craicoverflow:
Before merging @camilamacedo86 I think the mock file name should not have end with "_test", as Go will believe these are test files. It should just end with "_mock".

camilamacedo86:
@craicoverflow the changes requested were done. Have anything else that you would like to get done here before the merge or we can move forward?

craicoverflow:
> @craicoverflow the changes requested were done. Have anything else that you would like to get done here before the merge or we can move forward?

@camilamacedo86 that's great thanks. You can merge.

README.adoc
camilamacedo86: Cool to know. I will add it in the readme.

PR #58, opened 2019-02-20
feat(unit-tests): Increase db package code coverage to 86.7% (AEROGEAR-8555)

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8555

## What

Increased test coverage to 86.7% for the db package.

## Why

Code coverage stood at 67% for a critical package. This needed to be increased.

## How

Took the existing testing implementations for this package and converted them to table-driven tests, meaning more cases can easily be added to or removed from each test method.

## Verification Steps
 
1. From the root of the project, run the following command to test the `db` package:

```sh
go test ./pkg/db -cover -tags=integration
```

2. Check the output. It should say the code coverage achieved during this test. Did it get 86.7%?

```sh
$ go test ./pkg/db -cover -tags=integration
ok      github.com/aerogear/mobile-security-service/pkg/db      10.030s coverage: 86.7% of statements
```

3. Inspect the tests to ensure the conditions are correct.

## Checklist:

## Progress

- [x] Finished task

## Additional Notes

db go

### DIFF
--- a/pkg/db/db_test.go
+++ b/pkg/db/db_test.go
@@ -3,61 +3,133 @@
 package db
 
 import (
+	"database/sql"
 	"testing"
 
 	"github.com/aerogear/mobile-security-service/pkg/config"
-
 	_ "github.com/lib/pq"
 )
 
 func TestConnect(t *testing.T) {
 	config := config.Get()
-	dbConn, err := Connect(config.DB.ConnectionString, config.DB.MaxConnections)
 
-	if err != nil {
-		t.Errorf("Connect() returned an error: %v", err.Error())
+	type args struct {
+		connString     string
+		maxConnections int
 	}
+	tests := []struct {
+		name    string
+		args    args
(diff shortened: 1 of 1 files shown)


davidffrench:
Good work @craicoverflow . I would be interested to run the same cover on the other mobile service project file and see if that covers the same logic as ours.

craicoverflow:
> Good work @craicoverflow . I would be interested to run the same cover on the other mobile service project file and see if that covers the same logic as ours.

@davidffrench I tested the other mobile services project which is similar to this project, [aerogear-app-metrics], which has coverage of 81% on the `db` package.

For the conditions that we cannot reach in our test cases - these will fail in a real-life scenario when the actual code is incorrect. So I don't think we need to worry about this, it's just not possible to simulate these scenarios as the supplied arguments do not influence the outcome of these conditions.

PR #68, opened 2019-02-21
fix: sort on table

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8373

## What
Fix Sorting on Columns in Apps Table.
Previous Behaviour
- Sort only works once for each column and only ascending

## Why
Functionality wasn't working correctly 

## How
Add sortBy to redux state management
Update this when the sort is called so that the sortBy direction is reversed each time.

## Verification Steps

**Functionality**
Run the application
Click on the column header for each column multiple times
Sorting should work

**Run tests**
Do the above and then run
`npm test` - follow output to run all tests.

## Checklist:

## Progress

- [x] Functionality
- [ ] Serious Clean UP
- [ ] Testing

## Additional Notes

Pushing this as a draft pr as the functionality is now working but it needs a serious clean up!!
Lots of stuff in there that could be removed/cleaned but wanting to get some feedback on the general approach initially.


ui/src/actions/actions-ui.js
```diff
@@ -0,0 +1,16 @@
+import { SET_SORT, GET_SORT } from '../actions/types.js';
+
+export function reverseAppsTableSort (index, direction) {
+  const sortBy = { direction, index };
+  console.log('reverseAppsTableSort');
+  return {
+    type: SET_SORT,
```
damienomurchu: Just wondering if this should be `SORT_APPS` ?

ui/src/containers/AppsTableContainer.js
```diff
@@ -4,38 +4,49 @@ import AppsTable from '../components/AppsTable';
...
+      direction = SortByDirection.asc;
+    }
     this.setState({
       sortBy: {
         index,
         direction
-      },
+      }
+    });
+    this.setState({
```
damienomurchu: Should we manage this under the redux store?

laurafitzgerald:
> This looks good @laurafitzgerald. Just left some quick comments in the code after a quick scan of it (will take a deeper look in a bit). Verified the sorting works when I ran it locally too.
> 
> I see there's some state related to the table/ sorting still not managed through redux - should we fix that in this PR?

@damienomurchu yes we can definitely fix it here. This pr was a first attempt hack to get it all functioning. Definitely some improvements can be made, including the ones you commented above. Going at it again today with a fresh mind to try to clean it up. Will definitely ping you for review also when I have something new pushed. 😄

laurafitzgerald:
@damienomurchu I've pushed a clean up of the state management. There is a lot less flows to manage the same state as was managed before. I'm also considering that the reversal of the apps could all be done using the same action. Going to investigate that now but feel free to review what's there for now if you have some time.

laurafitzgerald:
Also I toyed with combining reducers at a stage where I had the reducer separated in the same way that the actions are seperated but I'm thinking now that we should have one actions file and one reducer file managing the state for each of the views. It's just a thought at this stage. I initially thought that separating the reducers for app data and ui state was the way to go but after some investigation I'm against that idea for this small app.

damienomurchu:
> I'm also considering that the reversal of the apps could all be done using the same action. 

Yes, I was thinking this too - one toggle action should suffice I think, wdyt?

damienomurchu:
> Also I toyed with combining reducers at a stage where I had the reducer separated in the same way that the actions are seperated but I'm thinking now that we should have one actions file and one reducer file managing the state for each of the views. It's just a thought at this stage. I initially thought that separating the reducers for app data and ui state was the way to go but after some investigation I'm against that idea for this small app.

Yes, lets not prematurely optimise. I think we can refactor easily enough if any research we do indicates we should structure the reducers and actions in a specific way - wdyt?

damienomurchu:
Great work @laurafitzgerald this looks good. One question I had when looking through the code - do we need `AppsTableContainer` or is it an unneeded abstraction and everything could be housed on `AppsTable` instead?

ui/src/reducers/index.js
```diff
@@ -1,13 +1,23 @@
-import { GET_APPS } from '../actions/types';
+import { GET_APPS, SET_APPS, SET_SORT, GET_SORT } from '../actions/types.js';
 
-const intialState = { apps: [] };
+const intialState = { apps: [], onSort: {} };
```
craicoverflow: Variable is spelled incorrectly. Should be `initialState`.

craicoverflow:
By the way this PR is still in Draft mode.

laurafitzgerald:
> Great work @laurafitzgerald this looks good. One question I had when looking through the code - do we need `AppsTableContainer` or is it an unneeded abstraction and everything could be housed on `AppsTable` instead?

Hi @damienomurchu I'm not sure about it but we did chat with @wei-lee around this topic in [this thread](https://groups.google.com/d/msg/aerogear/ltBjBo7Dsys/jTzTDgGvAAAJ) so I'm happy to take either approach. What I've done here is that I've put all the code dealing with state and logic in the container and passed that down to a UI rendering component. As you said though there may not be enough in the UI component to warrant this. I'm open to others opinions here.

laurafitzgerald:
> By the way this PR is still in Draft mode.

@craicoverflow I set it as draft mode for the first commit as the Redux stuff needed a serious refactor. I've done this now so does it make sense to change it from draft mode now?

craicoverflow:
> > By the way this PR is still in Draft mode.
> 
> @craicoverflow I set it as draft mode for the first commit as the Redux stuff needed a serious refactor. I've done this now so does it make sense to change it from draft mode now?

Hmm I'm not sure so. I guess it can be left :)

ui/src/actions/actions-ui.js
```diff
@@ -0,0 +1,18 @@
+import { REVERSE_SORT } from '../actions/types.js';
+import { SortByDirection } from '@patternfly/react-table';
+
+export function reverseAppsTableSort (direction, index, apps) {
```
craicoverflow: This can be made into an arrow function and can be changed to a const to make it immutable (which is what we want as we are not modifying existing states, we are creating a new one each time.

```suggestion
export const reverseAppsTableSort = (direction, index, apps) => {
```

laurafitzgerald:
The tests in the most recent commit will fail. Pushing for review/fresh set of eyes.

laurafitzgerald:
@damienomurchu the test is now working. I've pushed it up but will progress with squashing everything and this one should be ready for a final review soon!

laurafitzgerald:
@wei-lee thanks for the help with fixing the tests. Would you mind doing a quick review on this one now?

ui/src/containers/AppsTableContainer.js
```diff
@@ -1,57 +1,47 @@
...
     const sortedRows = this.props.apps.sort((a, b) => (a[index] < b[index] ? -1 : a[index] > b[index] ? 1 : 0));
-    this.setState({
-      sortBy: {
-        index,
-        direction
-      },
-      apps: direction === SortByDirection.asc ? sortedRows : sortedRows.reverse()
-    });
+    const sortedApps = this.props.sortBy.direction === SortByDirection.asc ? sortedRows : sortedRows.reverse();
+    this.props.reverseAppsTableSort(this.props.sortBy.direction, index, sortedApps);
```
wei-lee: while this does work, however, I think it will be better if the reverse part is removed to be part of the reducer. At the moment, if you read the `REVERSE_SORT` code in reducer, it doesn't actually do anything. People will confused why it is not doing the reverse there (at least I was). Normally speaking, in redux, all the data manipulation should happen there.

laurafitzgerald: @wei-lee do you mean line 20 should be moved out to the action. The logic that currently lives in the action reverses the SortByDirection value which is mapped to props.sortBy.direction. Should this also be done in the reducer?

ui/src/containers/AppsTableContainer.js
```diff
@@ -1,57 +1,47 @@
...
     super(props);
-    this.sortBy = {};
     this.onSort = this.onSort.bind(this);
     this.onRowClick = this.onRowClick.bind(this);
   }
-  componentDidMount () {
-    this.props.getApps();
-  }
   onRowClick (event, rowId, props) {
+    console.log('On row click called');
```
damienomurchu: Is this method needed? I don't think we use it or plan to use it

laurafitzgerald: @damienomurchu we'll use it later for navigation to the AppDetailedView, which should be covered in this [JIRA](https://issues.jboss.org/browse/AEROGEAR-8552)

laurafitzgerald: I can remove it for now if you like but it would get added back in later I imagine, unless we go with a different approach for navigating. I think though based on the way the table structure is wrapped up internally in the Patternfly table this is probably the easiest approach.

ui/src/containers/AppsTableContainer.js
damienomurchu: agree with @wei-lee, the `onSort()` should really just be making an action call to a reducer that looks after the data manipulation

ui/src/containers/AppsTableContainer.js
damienomurchu: @laurafitzgerald perhaps leave it in for now then, as I see what you mean. Let's see how we implement the routing/ redirect to the UI detail page when an app is clicked on

ui/package.json
```diff
@@ -29,12 +29,13 @@
     "not op_mini all"
   ],
   "devDependencies": {
+    "babel-core": "^7.0.0-bridge.0",
```
damienomurchu: Just a quick check - is this needed?

ui/src/containers/AppsTableContainer.js
laurafitzgerald: @damienomurchu 👍 I was thinking about taking that one next. Will resolve this conversation for now.

ui/package.json
laurafitzgerald: @damienomurchu good question, I added it as a dependency based on [this github issue](https://github.com/facebook/create-react-app/issues/5259) as I was seeing the error described a while back. I did a smoke test by removing it and doing a fresh npm install and I'm not seeing the error now when running the tests.  We could remove it and if the issue arrises we could add it back in. It's a fix on an npm issue and not a dependency we're using in the project. It's also possible that a change in the code since seeing the issue removed the cause of the issue but I don't recall at exactly what stage I saw it.

ui/src/containers/AppsTableContainer.js
laurafitzgerald: 👍 going to fix this now.

ui/package.json
laurafitzgerald: Yeah scratch that. I'm still seeing the error, so we need this dependency. 😂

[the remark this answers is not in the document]
laurafitzgerald: It uses babel core under the hood. This dependency is there to resolve an error faced. There was some discussion here about it also: https://github.com/aerogear/mobile-security-service/pull/68#discussion_r260366905

craicoverflow: Hi @laurafitzgerald if this is an issue with NPM then it's an issue with your environment and not the project itself. You should update your own NPM version and see if that works using `@babel/core`.

What version of NPM are you on? I am on version 6.8.0.

What is the error you get? I can't reproduce.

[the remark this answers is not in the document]
craicoverflow: I've replied to https://github.com/aerogear/mobile-security-service/pull/68#discussion_r260366905.

laurafitzgerald:
@wei-lee @craicoverflow @damienomurchu thanks for the comments. I've addressed those in the new commit. Would you mind reviewing those before I squash them.

PR #69, opened 2019-02-22
fix(cover) : Coveralls should not consider the helpers data which are useful for the tests. 

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8581

## What
Make cover test ignore the seed and fixture. This files should not be counted since they do not require unit_tests. 

## Why
For when we add new helpers data to be used in the tests the coverage does not decrease

## How
- Create the pkg helpers to keep the fixture and seeds there. In this way, the go test cover can ignore/exclude them since they should not have their own tests 
- Change the references in the imports
- Extract the seed func responsible to create the inserts for its own routine

## Verification Steps
- Check that the coveralls  increase since it is not checking the new helpers pkg
screenshot 2019-02-22 at 08 45 55

- All CI should finish successfully 

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/helpers/db_seed.go
+++ b/pkg/helpers/db_seed.go
@@ -1,15 +1,12 @@
-package test
+package helpers
 
 import (
 	"database/sql"
-
-	"github.com/aerogear/mobile-security-service/pkg/models"
-
 	"github.com/sirupsen/logrus"
 )
 
 // Seed the database with some sample data
-func seedDatabase(db *sql.DB) {
+func SeedDatabase(db *sql.DB) {
 	_, err := db.Exec(`INSERT INTO app
 			(id, app_id, app_name, deleted_at)
 		VALUES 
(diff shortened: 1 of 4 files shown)


damienomurchu:
@camilamacedo86 code changes in PR show the Coveralls change was effective - can I ask, where did you make the coveralls change? I can't see it documented in the JIRA, and think it would be good to do so, for anyone who needs to make a similar change in future - wdyt?

camilamacedo86:
Hi @damienomurchu,

The coveralls are made based on the output of the go command `go test -cover`.  According to its docs has no way to exclude or ignore the files/routine and the recommendation is to just passed to the cover the pkg which should be tested/covered.  

Currently, the command is already looking for the pkg. The issue was that we did not have a specific pkg for these routines. Is it make sense?

PR #74, opened 2019-02-28
feat(UI): adds Header component

damienomurchu:
## Motivation
Add header to landing page screen. AEROGEAR-8375

## What
Header for landing page ui screen

## Why
header

## How
Create header component with PF4 & add to Landing Page

## Verification Steps
- `npm start` 
- visually confirm header is rendered on landing page, and includes avatar icon, username, and dropdown with logout option
- verify dropdown toggles opens and closed
- verify dropdown closes after logout option selected
- run `npm test` & confirm all tests pass

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/ui/src/components/common/Header.css
+++ b/ui/src/components/common/Header.css
@@ -1,0 +1,8 @@
+.mssHeader {
+  background-color: black;
+}
+
+a {
+  text-decoration: none;
+  color: aliceblue;
+}
--- a/ui/src/components/common/Header.js
+++ b/ui/src/components/common/Header.js
@@ -1,0 +1,85 @@
+import React from 'react';
+
+import {
+  Dropdown,
+  DropdownToggle,
+  DropdownItem,
+  Page,
+  PageHeader,
+  Toolbar,
+  ToolbarGroup,
+  ToolbarItem
+} from '@patternfly/react-core';
+import { UserIcon } from '@patternfly/react-icons';
+import accessibleStyles from '@patternfly/patternfly/utilities/Accessibility/accessibility.css';
+import { css } from '@patternfly/react-styles';
+import './Header.css';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { toggleHeaderDropdown } from '../../actions/actions-ui';
+
+class Header extends React.Component {
+  onTitleClick = () => {};
+
+  onUserDropdownToggle = () => {
+    this.props.toggleHeaderDropdown();
+  };
+
+  onUserDropdownSelect = () => {
+    this.onUserDropdownToggle();
+  };
+
+  onLogoutUser = () => {
+    console.log('onLogoutUser()');
+  };
+
+  render () {
+    const logoProps = {
+      onClick: () => this.onTitleClick()
+    };
+
+    const toolbar = (
+      <Toolbar>
+        <ToolbarGroup className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnLg)}>
+          <UserIcon />
+          <ToolbarItem className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnMd)}>
+            <Dropdown
+              isPlain
+              position="right"
+              onSelect={this.onUserDropdownSelect}
+              isOpen={this.props.isUserDropdownOpen}
+              toggle={<DropdownToggle onToggle={this.onUserDropdownToggle}>{this.props.currentUser}</DropdownToggle>}
+              dropdownItems={[
+                <DropdownItem key="logout" component="button" href="#logout" onClick={this.onLogoutUser}>
+                  Log out
+                </DropdownItem>
+              ]}
+            />
+          </ToolbarItem>
+        </ToolbarGroup>
+      </Toolbar>
+    );
+
+    const Header = <PageHeader logo="MOBILE SECURITY SERVICE" logoProps={logoProps} toolbar={toolbar} />;
+
+    return (
+      <div className="mssHeader">
+        <Page header={Header} />
+      </div>
+    );
+  }
+}
+
+Header.propTypes = {
+  currentUser: PropTypes.string.isRequired,
+  isUserDropdownOpen: PropTypes.bool.isRequired
+};
+
+function mapStateToProps (state) {
+  return {
+    currentUser: state.currentUser,
+    isUserDropdownOpen: state.isUserDropdownOpen
+  };
+}
+
+export default connect(mapStateToProps, { toggleHeaderDropdown })(Header);
--- a/ui/src/reducers/index.js
+++ b/ui/src/reducers/index.js
@@ -1,18 +1,26 @@
-import { APPS_FAILURE, APPS_SUCCESS, APPS_REQUEST, REVERSE_SORT } from '../actions/types.js';
+import { APPS_FAILURE, APPS_SUCCESS, APPS_REQUEST, REVERSE_SORT, TOGGLE_HEADER_DROPDOWN } from '../actions/types.js';
+
 import { SortByDirection, sortable } from '@patternfly/react-table';
 
 const columns = [
   { title: 'App ID', transforms: [ sortable ] },
   { title: 'Deployed Versions', transforms: [ sortable ] },
   { title: 'Current Installs', transforms: [ sortable ] },
-  { title: 'Launches', transforms: [sortable] }
+  { title: 'Launches', transforms: [ sortable ] }
 ];
 
 const apps = { rows: [], data: {} };
 
 const sortBy = { direction: SortByDirection.asc, index: 0 };
 
-const initialState = { apps: apps, sortBy: sortBy, columns: columns, isAppsRequestFailed: false };
+const initialState = {
+  apps: apps,
+  sortBy: sortBy,
+  columns: columns,
+  isAppsRequestFailed: false,
+  currentUser: 'currentUser',
+  isUserDropdownOpen: false
+};
 
 export default (state = initialState, action) => {
   switch (action.type) {
@@ -37,7 +45,7 @@ export default (state = initialState, action) => {
       };
     case APPS_SUCCESS:
       var fetchedApps = [];
-      action.result.forEach(app => {
+      action.result.forEach((app) => {
         var temp = [];
         temp[0] = app.appName;
         temp[1] = app.numOfDeployedVersions;
@@ -56,6 +64,12 @@ export default (state = initialState, action) => {
       return {
         ...state,
         isAppsRequestFailed: true
+      };
+    case TOGGLE_HEADER_DROPDOWN:
+      const isUserDropdownOpen = state.isUserDropdownOpen;
+      return {
+        ...state,
+        isUserDropdownOpen: !isUserDropdownOpen
       };
     default:
       return state;
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -3503,7 +3503,8 @@
             },
             "ansi-regex": {
               "version": "2.1.1",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -3521,11 +3522,13 @@
             },
             "balanced-match": {
               "version": "1.0.0",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
+              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
@@ -3538,15 +3541,18 @@
             },
             "code-point-at": {
               "version": "1.1.0",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "concat-map": {
               "version": "0.0.1",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "console-control-strings": {
               "version": "1.1.0",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "core-util-is": {
               "version": "1.0.2",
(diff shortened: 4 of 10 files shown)


damienomurchu:
@laurafitzgerald just pinging you on this to get your eyes as I try construct a header with PF4 components

ui/package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@patternfly/react-core": "2.1.4",
     "@patternfly/react-table": "1.0.7",
+    "prop-types": "^15.7.2",
```
laurafitzgerald: @damienomurchu I'm also using Prop Types here: https://github.com/aerogear/mobile-security-service/blob/master/ui/src/containers/AppsTableContainer.js#L3 but I didn't need to add it as a dependency. It may already be there as part of the React or Redux dependencies?

ui/src/components/common/Header.js
```diff
@@ -0,0 +1,105 @@
...
+
+    /* const logoProps = {
+      href: 'https://patternfly.org',
+      onClick: () => console.log('clicked logo'),
+      target: '_blank'
+    }; */
+
+
+    const toolbar = (
+      <>
```
laurafitzgerald: I'm not familiar with the use of `<>`. Is this a PF thing?

damienomurchu: A React thing actually - it's shorthand for [`<React.Fragment>`](https://reactjs.org/docs/fragments.html). Must see if it's actually needed here though

damienomurchu:
thanks for the feedback @laurafitzgerald - if you get a chance could you take a look at the rendered header & see if it's what we were thinking of?

ui/src/components/common/Header.js
laurafitzgerald: @damienomurchu given that you both open and close the Fragment with Toolbar, I'm pretty sure that you don't need it.

laurafitzgerald:
> thanks for the feedback @laurafitzgerald - if you get a chance could you take a look at the rendered header & see if it's what we were thinking of?

@damienomurchu yes I would say it's on the right track. It's starting to look similar to the one in MDC. The black background looks well.

ui/src/components/common/Header.js
```diff
@@ -0,0 +1,105 @@
...
+              />
+            </ToolbarItem>
+          </ToolbarGroup>
+        </Toolbar>
+      </>
+    );
+
+    const Header = (
+      <PageHeader
+        logo="Mobile Developer Console" logoProps={logoProps} toolbar={toolbar} />
```
laurafitzgerald:
```suggested
logo="Mobile Security Service" logoProps={logoProps} toolbar={toolbar} />
```

damienomurchu:
@laurafitzgerald thanks - just looking at the flat, think just an avatar/ user icon next to the dropdown is required, and will add this. 

Will add some tests & ping you for re-review once I do so and remove the draft label

damienomurchu:
Removing this from draft status & reviews welcome

laurafitzgerald:
@damienomurchu 👀

ui/package.json
damienomurchu: removed now :+1:

damienomurchu:
@laurafitzgerald thanks for your eyes. As discussed need to wire this up to the redux store now, and will ping you for review then.

ui/src/components/common/Header.css
```diff
@@ -0,0 +1,3 @@
+.mssHeader {
```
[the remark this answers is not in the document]
damienomurchu: Good question, and given that this was only one line of CSS, I was even considering adding it via an inline style to the component, and not having a seperate css file for the Header. I think can chat about this off-PR & not let it block the PR as if we can refactor easily even if this is merged in as it is above

ui/src/components/common/Header.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  onTitleClick = () => {
+    const { history } = this.props;
+    history.push('/');
+  };
+
+  onUserDropdownToggle(isUserDropdownOpen) {
+    this.setState({
+      isUserDropdownOpen
```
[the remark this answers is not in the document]
damienomurchu: Added this to redux now. FWIW I don't like the idea of mixing UI state between redux & using setState, and think Redux should be the single place we manage state. We can discuss this off-PR though I think :)

ui/src/components/common/Header.js
```diff
@@ -0,0 +1,93 @@
...
+                  Log out
+                </DropdownItem>
+              ]}
+            />
+          </ToolbarItem>
+        </ToolbarGroup>
+      </Toolbar>
+    );
+
+    const Header = <PageHeader logo="Mobile Security Service" logoProps={logoProps} toolbar={toolbar} />;
```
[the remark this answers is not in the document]
damienomurchu: @laurafitzgerald Yes, this is how the text is rendered in PF4. Will correct case to all uppercase

ui/src/components/common/Header.js
```diff
@@ -0,0 +1,93 @@
...
+    this.onTitleClick = this.onTitleClick.bind(this);
+    //this.onLogoutUser = this.onLogoutUser.bind(this);
+
+    this.onUserDropdownToggle = this.onUserDropdownToggle.bind(this);
+    this.onUserDropdownSelect = this.onUserDropdownSelect.bind(this);
+  }
+
+  onTitleClick = () => {
+    const { history } = this.props;
+    history.push('/');
```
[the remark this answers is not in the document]
damienomurchu: @laurafitzgerald I agree, and have removed this for now. I think we should discuss how we intend to manage page history in our app (or if we choose to manage it at all)

ui/src/components/common/Header.js
```diff
@@ -14,43 +14,28 @@ import { UserIcon } from '@patternfly/react-icons';
...
-    };
-  }
-
-  onTitleClick = () => {
-    console.log('onTitleClick()');
-    // const { history } = this.props;
-    // history.push('/');
+  onUserDropdownToggle = () => {
+    const isUserDropdownOpen = this.props.isUserDropdownOpen;
+    this.props.toggleHeaderDropdown(isUserDropdownOpen);
```
laurafitzgerald: @damienomurchu I did the same when I initially set up the `isappsrequestfailed` where I passed the value through the component, action and reducer. As you are reversing it in the reducer you don't need to pass the value down.

ui/src/components/common/Header.css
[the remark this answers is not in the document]
laurafitzgerald: Yes agreed, it's probably a bigger conversation. Happy to move the convo elsewhere to unblock this pr.

ui/src/components/common/Header.js
damienomurchu: good point @laurafitzgerald :+1:

ui/src/components/common/Header.js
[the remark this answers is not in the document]
laurafitzgerald: 👍 Could you propose a new task via the team channel? We could address it as part of either UI story in my opinion.

laurafitzgerald:
@damienomurchu thanks for pushing the changes. There's one console.log in the reducer which can be removed. Also I'm not sure about the `MOBILE SECURITY SERVICE` being rendered in link format. i.e. in blue and underline. It's deviating from the MDC layout and the flats somewhat. Could be worth a chat with Chris about it if it's the only way to do it in PF4.

damienomurchu:
> @damienomurchu thanks for pushing the changes. There's one console.log in the reducer which can be removed. Also I'm not sure about the `MOBILE SECURITY SERVICE` being rendered in link format. i.e. in blue and underline. It's deviating from the MDC layout and the flats somewhat. Could be worth a chat with Chris about it if it's the only way to do it in PF4.

Console.log removed now :)

Re `MOBILE SECURITY SERVICE` being rendered in link format, my line of thought is that this would enable the user to always navigate back to the root page of the UI and supplement our app navigation. Wdyt? Perhaps we can discuss this & the navigation in general in the UI?

laurafitzgerald:
> > @damienomurchu thanks for pushing the changes. There's one console.log in the reducer which can be removed. Also I'm not sure about the `MOBILE SECURITY SERVICE` being rendered in link format. i.e. in blue and underline. It's deviating from the MDC layout and the flats somewhat. Could be worth a chat with Chris about it if it's the only way to do it in PF4.
> 
> Console.log removed now :)
> 
> Re `MOBILE SECURITY SERVICE` being rendered in link format, my line of thought is that this would enable the user to always navigate back to the root page of the UI and supplement our app navigation. Wdyt? Perhaps we can discuss this & the navigation in general in the UI?

Yes, I think it should be a home link. but I don't think it should behave like a normal link. i.e. it shouldn't be Blue and underlined. I think it should appear in the same way it does in the MDC and in the flats.

damienomurchu:
@laurafitzgerald agree, that can be fixed :+1:

ui/src/components/common/Header.js
```diff
@@ -0,0 +1,84 @@
...
+import { css } from '@patternfly/react-styles';
+import './Header.css';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { toggleHeaderDropdown } from '../../actions/actions-ui';
+
+class Header extends React.Component {
+  onTitleClick = () => {};
+
+  onUserDropdownToggle = () => {
```
laurafitzgerald: @damienomurchu Looks like this method is a duplicate of `onUserDropdownSelect()`

ui/src/components/common/Header.js
```diff
@@ -0,0 +1,84 @@
...
+      <Toolbar>
+        <ToolbarGroup className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnLg)}>
+          <UserIcon />
+          <ToolbarItem className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnMd)}>
+            <Dropdown
+              isPlain
+              position="right"
+              onSelect={this.onUserDropdownSelect}
+              isOpen={this.props.isUserDropdownOpen}
+              toggle={<DropdownToggle onToggle={this.onUserDropdownToggle}>currentUserName</DropdownToggle>}
```
laurafitzgerald: @damienomurchu we'll probably do the implemenation later, but could you move the `currentUserName` hardcoded value to Redux state?

ui/src/components/common/Header.js
damienomurchu: @laurafitzgerald both needed - `onUserDropdownToggle` will be triggered when the unexpanded dropdown is clicked on, and `onUserDropdownSelect` is needed to close the dropdown when a dropdown item has been selected

ui/src/components/common/Header.js
```diff
@@ -0,0 +1,84 @@
...
+    const toolbar = (
+      <Toolbar>
+        <ToolbarGroup className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnLg)}>
+          <UserIcon />
+          <ToolbarItem className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnMd)}>
+            <Dropdown
+              isPlain
+              position="right"
+              onSelect={this.onUserDropdownSelect}
+              isOpen={this.props.isUserDropdownOpen}
```
laurafitzgerald: Hey, I was thinking about this one yesterday. Because both methods do the same thing, you could pass the same method to the isOpen for the Dropdown component and to the onToggle of the DropdownToggle component. wdyt?

damienomurchu: @laurafitzgerald was thinking something similar, to reduce code duplication & make any future changes easier :+1:

damienomurchu:
@laurafitzgerald good point - will add to Detailed Apps UI page. Re redux testing, I was thinking about that too, and will do it in another PR I think - will also ping you probably re redux testing as its not something I've done any of yet :)

laurafitzgerald:
@damienomurchu cool. There's some example of how the tests can be done in my [previous pr](https://github.com/aerogear/mobile-security-service/pull/75/files#diff-8ce8f218c3acf4b3d431453881378f99)

damienomurchu:
@laurafitzgerald added redux test now for dropdown state if you could re-review (& approve if happy). Don't see need to add redux test for currentUsername until we implement it properly by the way, so have not added any redux test for that

ui/src/components/common/Header.js
laurafitzgerald: @damienomurchu This one still needs to be done?

damienomurchu:
Accidentally broke tests before, but fixed now

ui/src/reducers/index.js
```diff
@@ -35,7 +43,7 @@ export default (state = initialState, action) => {
       };
     case APPS_SUCCESS:
       var fetchedApps = [];
-      action.result.forEach(app => {
+      action.result.forEach((app) => {
```
laurafitzgerald: What's the reason for this change?

damienomurchu: just the auto-linter, no functional change

ui/src/components/common/Header.js
damienomurchu: @laurafitzgerald misread your previous comment - isOpen is just a boolean value, and onSelect is the event handler.

damienomurchu:
@laurafitzgerald thanks - responded inline. If you're happy let me know & I'll merge this in

ui/src/components/common/Header.js
laurafitzgerald: Sorry I mean onSelect and onToggle

damienomurchu: onToggle opens the dropdown, and onSelect closes it. I think its probably best to keep both methods to keep table implementation as close to standard configuration as possible though.

laurafitzgerald: I disagree. We aren't changing the the table implementation, only the methods that are passed to each of the required props. But it's a small detail, as you wish.

laurafitzgerald:
@damienomurchu LGTM

PR #76, opened 2019-03-01
feat(init): Create /init endpoint for POST (AEROGEAR-8360)

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8360

## What

Added init endpoint.

## Why

Required to return information to the app.

## Verification Steps

Send a POST request to `/api/init` with the following data:

```json
{
	"deviceId": "d1895cc1-28d7-4283-932d-8bcab9e4a469",
	"deviceVersion": "3.4",
	"version": "3.10",
	"deviceType": "Android",
	"appId": "com.aerogear.testapp"
}
```

You should receive the init data which looks something like this:

```json
{
    "id": "4ca6376c-61cc-4c5f-8a47-f17dd8b9e624",
    "version": "3.10",
    "appId": "com.aerogear.testapp",
    "disabled": true,
    "disabledMessage": "DISABLED!!"
}
```

Change the version in the original POST request. After POSTing again you should still get back data. A new version was created in the database.

Check the version table for that version. Is the `num_of_app_launches` counter increasing every time you make a request?

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

This task is not finished. The repository unit tests have not been completed. Please review what is there.


Gopkg.toml
```diff
@@ -13,6 +13,11 @@
 [[constraint]]
   name = "github.com/joho/godotenv"
   version = "1.3.0"
+
 [[constraint]]
   name = "gopkg.in/DATA-DOG/go-sqlmock.v1"
   version = "1.3.2"
+
+[[constraint]]
+  name = "github.com/satori/go.uuid"
```
austincunningham: Use the Google UUID checker

austincunningham: looks like its just left over in the toml file

craicoverflow: Yes this needs to be removed. Where is [google/uuid](https://github.com/google/uuid) though @austincunningham?

pkg/web/apps/apps_http_handler_test.go
```diff
@@ -131,7 +131,9 @@ func Test_httpHandler_GetAppByID(t *testing.T) {
 				t.Errorf("HTTPHandler.GetAppByID() statusCode = %v, wantCode = %v", rec.Code, tt.wantCode)
 			}
 			if strings.TrimSpace(rec.Body.String()) != tt.want {
-				t.Errorf("httpHandler.GetAppByID() got %v, want %v", strings.TrimSpace(rec.Body.String()), tt.want)
+				t.Log(strings.TrimSpace(rec.Body.String()))
+				t.Log(tt.want)
```
camilamacedo86: Is the log return an error? Shows that this point here need bee fixed.

pkg/models/init_client.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+// InitClient model
+// swagger:model Init
+type InitClient struct {
```
davidffrench: what are the pros/cons of this new model over the version model?

pkg/web/apps/apps_service.go
```diff
@@ -53,3 +55,77 @@ func (a *appsService) GetAppByID(id string) (*models.App, error) {
...
+		return nil, err
+	}
+
+	// If the version does not exist, create it
+	if err == models.ErrNotFound {
+		// Create new uuid for our new app version
+		versionUUID := uuid.New()
+
+		version = &models.Version{
+			ID:               versionUUID.String(),
```
camilamacedo86: Here it can be uuid.New().String directly.

pkg/web/apps/apps_service.go
```diff
@@ -53,3 +55,77 @@ func (a *appsService) GetAppByID(id string) (*models.App, error) {
...
+	// If we can't find the device by version and app ID
+	if _, err = a.repository.GetDeviceByVersionAndAppID(version.Version, deviceInfo.AppID); err == models.ErrNotFound {
+
+		// If we can't find the device by device ID and app ID
+		if _, err := a.repository.GetDeviceByDeviceIDAndAppID(deviceInfo.DeviceID, deviceInfo.AppID); err == models.ErrNotFound {
+
+			id := uuid.New()
+
+			// Build a new device to save to the database
+			device := models.Device{
```
camilamacedo86: It could be in the device model as a constructor method.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -128,5 +130,154 @@ func (a *appsPostgreSQLRepository) GetAppByID(ID string) (*models.App, error) {
...
+func (a *appsPostgreSQLRepository) GetDeviceByDeviceID(deviceID string) (*models.Device, error) {
+	query := `
+	SELECT id,version_id,app_id,device_id,device_type,device_version
+	FROM device
+	WHERE device_id = $1;`
+
+	row := a.db.QueryRow(query, deviceID)
+
+	var device models.Device
+	if err := row.Scan(&device.ID, &device.VersionID, &device.AppID, &device.DeviceID, &device.DeviceType, &device.DeviceVersion); err != nil {
```
davidffrench: There are some differences in the formatting of our error handling between different functions.

This function includes the row.Scan with the err conditional and is using a switch statement for the sql.ErrNoRows error check. The GetDeviceByDeviceIDAndAppID function has the QueryRow, Scan and err conditional on the same line and is not using a switch statement for the sql.ErrNoRows check. The other functions are using some version of the above aswell.

I don't have a preference on which one is chosen, the main goal is consistency. Even though it is a small thing.

pkg/web/apps/apps_service.go
```diff
@@ -53,3 +55,77 @@ func (a *appsService) GetAppByID(id string) (*models.App, error) {
...
+				Version:       version.Version,
+				AppID:         deviceInfo.AppID,
+				DeviceID:      deviceInfo.DeviceID,
+				DeviceVersion: deviceInfo.DeviceVersion,
+				DeviceType:    deviceInfo.DeviceType,
+			}
+
+			// Could not insert the device
+			if err := a.repository.CreateDevice(&device); err != nil {
+				log.Error(err)
```
camilamacedo86: Should return error.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -128,5 +130,154 @@ func (a *appsPostgreSQLRepository) GetAppByID(ID string) (*models.App, error) {
...
+	}
+
+	return &device, nil
+}
+
+// GetVersionByAppIDAndVersion gets a version by its app ID and version number
+func (a *appsPostgreSQLRepository) GetVersionByAppIDAndVersion(appID string, versionNumber string) (*models.Version, error) {
+	version := models.Version{}
+	var disabledMessage sql.NullString
+	var lastLaunchedAt sql.NullString
```
davidffrench: If a version exists, this `lastLaunchedAt` will be set. There should never be a scenario where it is null in the database in my opinion.

pkg/web/apps/apps_service_test.go
```diff
@@ -85,3 +87,238 @@ func Test_appsService_GetAppByID(t *testing.T) {
...
+		wantErr error
+	}{
+		{
+			name:    "InitClient() should return ErrNotFound when app not found",
+			wantErr: models.ErrNotFound,
+			want:    nil,
+			args: args{
+				deviceInfo: &models.Device{
+					ID:            uuid.New().String(),
+					VersionID:     uuid.New().String(),
```
camilamacedo86: Use the fixture to create mock of the models/structures

pkg/web/apps/apps_psql_repository.go
```diff
@@ -128,5 +130,154 @@ func (a *appsPostgreSQLRepository) GetAppByID(ID string) (*models.App, error) {
...
+			return nil, models.ErrNotFound
+		}
+		return nil, models.ErrInternalServerError
+	}
+
+	return &device, nil
+}
+
+// GetAppByID retrieves an app by id from the database
+func (a *appsPostgreSQLRepository) GetAppByAppID(appID string) (*models.App, error) {
```
davidffrench: Small change, I would move this function to below the GetAppByID function.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -128,5 +130,154 @@ func (a *appsPostgreSQLRepository) GetAppByID(ID string) (*models.App, error) {
...
+		return nil, models.ErrInternalServerError
+	}
+
+	return &app, nil
+
+}
+
+// InsertVersionOrUpdateNumOfAppLaunches creates a new version row
+// or increments the num_of_app_launches counter if the version already exists
+func (a *appsPostgreSQLRepository) InsertVersionOrUpdateNumOfAppLaunches(version *models.Version) error {
```
davidffrench: The name currently does not capture that last_launched_at is updated as well and it would become quite long if that is added as well. This comment is not very helpful :-D

pkg/web/apps/apps_service_test.go
```diff
@@ -85,3 +87,238 @@ func Test_appsService_GetAppByID(t *testing.T) {
...
+					DeviceType:    "Android",
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			mockedRepository := &RepositoryMock{
```
camilamacedo86: Use the mock repository for success and error on top of it. Check how is implemented the other tests where the mockRepo is passed as an argument. For some specific scenario which is valid just here, you can create a mock locally just implementing the functions which are used at this test.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -128,5 +130,154 @@ func (a *appsPostgreSQLRepository) GetAppByID(ID string) (*models.App, error) {
...
+	sqlStatement := `
+		INSERT INTO version(id, version, app_id, disabled, disabled_message, num_of_app_launches, last_launched_at)
+		VALUES($1, $2, $3, $4, $5, $6, NOW())
+		ON CONFLICT (id)
+		DO UPDATE
+		SET num_of_app_launches = $6,
+		last_launched_at = NOW();`
```
davidffrench: According to the docs, this uses the current local time[1]. It might be good to make sure this is always UTC in my opinion.

[1] https://golang.org/pkg/time/#Now

pkg/web/initclient/init_http_handler.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if err != nil {
+		return httperrors.GetHTTPResponseFromErr(c, err)
+	}
+
+	return c.JSON(http.StatusOK, initResponse)
+}
+
+// Request body must have version, appId and deviceId
+func validateInitRequestBody(deviceInfo *models.Device) error {
```
camilamacedo86: It is not necessary, check in the updated versions that you can transform a JSON to a struct.

pkg/web/initclient/init_http_handler_test.go
```diff
@@ -0,0 +1,125 @@
...
+					Version:       "9.1.1",
+					DeviceType:    "Android",
+				},
+			},
+			wantStatusCode: 400,
+		},
+		{
+			name: "A 400 Bad Request should be returned when request body is missing version",
+			args: args{
+				device: models.Device{
```
camilamacedo86: As described above use the fixture to mock the struct.

pkg/web/router/router.go
```diff
@@ -125,6 +128,5 @@ func SetAppRoutes(r *echo.Group, appsHandler apps.HTTPHandler) {
 	//     description: Invalid appId supplied
 	//   404:
 	//     description: App not found
-	r.POST("/init", appsHandler.GetApps) // TODO: Implement the func which will be used here and its call
-
```
camilamacedo86: - Here shows are missing updating the swagger doc comments and re-generated the swagger API. You can use the make test,

pkg/web/apps/apps_service.go
```diff
@@ -53,3 +55,77 @@ func (a *appsService) GetAppByID(id string) (*models.App, error) {
...
+	version.NumOfAppLaunches++
+
+	// Update the existing version or create a new one
+	if err := a.repository.InsertVersionOrUpdateNumOfAppLaunches(version); err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	// If we can't find the device by version and app ID
+	if _, err = a.repository.GetDeviceByVersionAndAppID(version.Version, deviceInfo.AppID); err == models.ErrNotFound {
```
davidffrench: The logic for these device calls should be changed. The flow should be as follows:

device logic

pkg/web/apps/apps_psql_repository.go
```diff
@@ -128,5 +130,154 @@ func (a *appsPostgreSQLRepository) GetAppByID(ID string) (*models.App, error) {
 	}
 
 	return &app, nil
+}
+
+// GetDeviceByDeviceID gets a device by its Device ID
+func (a *appsPostgreSQLRepository) GetDeviceByDeviceID(deviceID string) (*models.Device, error) {
```
davidffrench: Is this function currently used anywhere?

pkg/web/apps/apps_http_handler_test.go
craicoverflow: Actually yeah I forgot about this. This test stopped working because of changes that happened outside of this task when they were merged in. It requires a sizeable amount of changes to get it working again so it should be in a separate task.

pkg/models/init_client.go
craicoverflow: The response isn't returning a full `Version` model. One reason for having a model is that it can be changed from a central location instead of within the endpoint.

The main reason though is how would we return a `Version` with only a select number of its properties? If there are properties in the `Version` model we don't want we would probably have to do something like:

```go
verson.PropOne = nil
version.PropThree = nil
```

I'm open to feedback on this, above way could be better but I think it's less clean.

pkg/web/apps/apps_psql_repository.go
craicoverflow: I will have a think for names. If you can think of any too that would be good - hard to find one :dagger:

pkg/models/init_client.go
davidffrench: Do you see any cons with using the existing Version model with `omitempty` on all the unwanted properties?

pkg/web/apps/apps_psql_repository.go
davidffrench: I was initially thinking of `UpsertVersion`, but this function is not that generic as we do not want to increment `num_of_app_launches` or set `last_launched_at` or any other update. It is only a small thing, so maybe something like `UpsertVersionWithAppLaunchesAndLastLaunched`.

pkg/web/apps/apps_psql_repository.go
craicoverflow: @davidffrench This is the PostgreSQL `NOW()` function.
http://www.postgresqltutorial.com/postgresql-now/

Actually now that I think about it - we have two TIMESTAMP columns - `deleted_at` and `last_launched_at`. Neither store a timestamp **without time zone**.

Is this what we want? From PostgreSQL [documentation](https://www.postgresql.org/docs/9.1/datatype-datetime.html):

> For timestamp with time zone, the internally stored value is always in UTC (Universal Coordinated Time, traditionally known as Greenwich Mean Time, GMT). An input value that has an explicit time zone specified is converted to UTC using the appropriate offset for that time zone. If no time zone is stated in the input string, then it is assumed to be in the time zone indicated by the system's timezone parameter, and is converted to UTC using the offset for the timezone zone.

How will the current implementation work in practice? Would it be better to store time zone information as it defaults to UTC? This probably requires a separate discussion/investigation - but its still relevant enough to be initiated here.

ERD

davidffrench: Good information. If we are always defaulting to the server time or UTC, there is no need to store the timezone. For now, I believe we can leave the fields as `TIMESTAMP WITHOUT TIMEZONE` and just use `NOW()` to store the server time. This will always be consistent.

pkg/web/apps/apps_service.go
craicoverflow: @davidffrench The version ID is not currently being sent in the `/init` POST,  so the init device and db device cannot be compared on this property. Am I missing something?

pkg/web/apps/apps_psql_repository.go
craicoverflow: Doesn't seem to be. Dunno why it is here :grinning:

pkg/web/apps/apps_service.go
davidffrench: it is not sent in the `/init` endpoint but is included in the version retrieved from the database. The property called `version` will have a `id` property you can check against.

pkg/web/apps/apps_psql_repository.go
camilamacedo86: Also, you can do it by the code using the time.now() function. See [here](https://github.com/aerogear/mobile-security-service/blob/master/pkg/web/apps/apps_psql_repository.go#L166)

pkg/web/apps/apps_psql_repository.go
craicoverflow: I agree. I will update the ones involved in this PR. There are probably inconsistencies elsewhere which I will not touch here.

pkg/web/apps/apps_http_handler_test.go
camilamacedo86: Hi, @craicoverflow  I am afraid that it is not a good scenario for another task. It is the kind of thing that still lost with time. You could divide here the tasks, in fact, you are doing 2 jiras in one PR. 

IMHO, leave logs and outputs in the code is not a good idea and make easier we forget this and do not come back to fix it.

pkg/helpers/fixtures.go
```diff
@@ -42,39 +43,57 @@ func GetMockApp() *models.App {
...
 //Get version strut for the disable all.
 func GetMockAppVersionForDisableAll() models.Version {
 	return models.Version{
-			DisabledMessage:      "Please contact an administrator",
-		}
+		DisabledMessage: "Please contact an administrator",
+	}
+}
+
+func GetMockDevice() *models.Device {
```
camilamacedo86: 👍 🥇 Excellent work here !!

pkg/web/apps/apps_psql_repository.go
camilamacedo86: I believe that it is duplicated. Please, @craicoverflow could you rebase your PR with the master and check if it is already implemented?

pkg/web/apps/apps_service_test.go
craicoverflow: Hi @camilamacedo86 I think I need to have it in here. All of these functions are in use in this test.

pkg/web/initclient/init_http_handler.go
craicoverflow: I didn't implement this, but the purpose of it is not to transform JSON to a struct. It is checking if the device model is missing properties and returns an error if so. Come to think of it, this may be better in the model.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -117,7 +119,131 @@ func (a *appsPostgreSQLRepository) GetAppByID(ID string) (*models.App, error) {
...
+
+	return nil
+}
+
+// InsertDeviceOrUpdateVersionID creates a new device row in the device table
+func (a *appsPostgreSQLRepository) InsertDeviceOrUpdateVersionID(device models.Device) error {
+	sqlStatement := `
+		INSERT INTO device(id,version_id,app_id,device_id,device_type,device_version)
+		VALUES($1, $2, $3, $4, $5, $6)
+		ON CONFLICT (id)
```
camilamacedo86: @craicoverflow How it will works if one device has more than one app/version installed? 
I'd like to suggest you do the implementation in the service and not in the repo layer as the following for example.

1. Check if has a row with the device and version 
2. If not create/insert
3. If yes, them update the totals in the version table

c/c @davidffrench

pkg/web/initclient/init_http_handler_test.go
craicoverflow: Using the mock fixture is not suitable here. This is simulating many different scenarios. As you can see the first test case has a device with a missing DeviceID property. Each of the other test cases do something similar expect with slightly different property changes. 

The fixtures only return valid mock data. We could create a new fixture method for each test case but then we are still writing the same amount of code except in a different file.

pkg/web/apps/apps_repository.go
```diff
@@ -11,4 +11,10 @@ type Repository interface {
 	GetAppVersionsByAppID(ID string) (*[]models.Version, error)
 	UpdateAppVersions(versions []models.Version) error
 	DisableAllAppVersionsByAppID(appID string, message string) error
+	GetVersionByAppIDAndVersion(appID string, versionNumber string) (*models.Version, error)
+	GetDeviceByDeviceIDAndAppID(deviceID string, appID string) (*models.Device, error)
+	GetDeviceByVersionAndAppID(versionID string, appID string) (*models.Device, error)
+	GetAppByAppID(appID string) (*models.App, error)
+	InsertVersionOrUpdateNumOfAppLaunches(version *models.Version) error
+	InsertDeviceOrUpdateVersionID(device models.Device) error
```
camilamacedo86: @craicoverflow it shows not be a good practice have a method/func which does more than one action. I mean you are losing cohesion when you do a func as InsertOrUpdate. I'd like to recommend you change it anyway.

craicoverflow: Hi @camilamacedo86 I would normally agree but this is done with a SQL feature known as `UPSERT`. Therefore it is really doing one action.

https://wiki.postgresql.org/wiki/UPSERT

pkg/web/apps/apps_service.go
```diff
@@ -79,3 +81,103 @@ func (a *appsService) DisableAllAppVersionsByAppID(id string, message string) er
...
+
+		isUpdated = true
+	}
+
+	return isUpdated
+}
+
+// If the Device.DeviceVersion property is different to @var deviceVersion,
+// update it to match, returning a bool to indicate if it was updated
+func updateDeviceDeviceVersion(device *models.Device, deviceVersion string) bool {
```
camilamacedo86: Update device device?

camilamacedo86: Why it is required?

pkg/web/apps/apps_service.go
```diff
@@ -79,3 +81,103 @@ func (a *appsService) DisableAllAppVersionsByAppID(id string, message string) er
...
+		device.DeviceVersion = deviceVersion
+
+		isUpdated = true
+	}
+
+	return isUpdated
+}
+
+// Build a new Device model from a combination of Device and Version model properties.
+func newDeviceFromDeviceAndVersion(d models.Device, v models.Version) *models.Device {
```
camilamacedo86: Who knows more one entity is the proper entity. I'd like to recommend you add in the models its constructors.In this way, in the model.device you will have a constructor as in the version you will have another. Also, adding the constructors in the models allow it to be used in other places.

pkg/web/apps/apps_service_test.go
camilamacedo86: Hi @craicoverflow, the file already has the mocks as global for global use. See [here](https://github.com/aerogear/mobile-security-service/blob/master/pkg/web/apps/apps_service_test.go#L12) 
[here](https://github.com/aerogear/mobile-security-service/blob/master/pkg/web/apps/apps_service_test.go#L44) the global ones used for all tests implemented in this file. 

Also, look [here](https://github.com/aerogear/mobile-security-service/blob/master/pkg/web/apps/apps_service_test.go#L278) to get an example of how to do when the global ones are not enough and is required do more and specifics checks.

pkg/web/initclient/init_http_handler.go
```diff
@@ -0,0 +1,52 @@
...
+	// Check the request body is valid
+	if err := deviceInfo.ValidateInitBody(); err != nil {
+		return httperrors.BadRequest(c, err.Error())
+	}
+
+	initResponse, err := h.appsService.InitClientApp(deviceInfo)
+
+	// If no app has been found in the database, return a bad request to the client
+	if err == models.ErrNotFound {
+		return httperrors.BadRequest(c, "No bound app found for the sent App ID")
```
camilamacedo86: Is not it missing the same validation for an invalid version?

pkg/web/initclient/init_http_handler_test.go
```diff
@@ -0,0 +1,125 @@
...
+	tests := []struct {
+		name           string
+		fields         fields
+		args           args
+		wantStatusCode int
+	}{
+		{
+			name: "A 400 Bad Request should be returned when request body is missing device ID",
+			args: args{
+				device: models.Device{
```
camilamacedo86: Use the helper to get the mode mocked.

pkg/web/apps/apps_psql_repository.go
camilamacedo86: @davidffrench then IHMO it is the wrong layer. The logic over what should be done is part of the service and not the repo. Following my suggestion

apps_service.go
```go
func (a *appsService) InitClientApp(deviceInfo *models.Device) (*models.Version, error) {
	
	version, err := a.repository.GetVersionByAppIDAndVersion(deviceInfo) 

	// If any error other Not Found error occurred, return
	if err != nil && err != models.ErrNotFound {
		return nil, err
	}

	device, err := a.repository.GetDeviceByVersionIDandAppID(deviceInfo)

	// If is a new installation create the device and update the totals
	if err != nil && err != models.ErrNotFound {
		device, err := a.repository.CreateDevice(deviceInfo)
		if err != nil {
			return err
		}
		a.repository.IncrementVersionTotals(version, false) 
	}
	//If is not a new installation then just update the totals 
	a.repository.IncrementVersionTotals(version, false)	
	return version, nil
}

func (a *appsService) cleanVersionResponseForInitCall(version *models.Version) (*models.Version, error) {
	//TODO: Remove from the answer the totals before return this information should not be sent to the mobile device 
}

```

Repository
```go
func (a *appsPostgreSQLRepository) IncrementVersionTotals(versionInfo models.Version, isNewInstall bool) error {

	if isNewInstall {
		//TODO: It will update the TOTALS in the version table adding +1 for new install and launches
	}


	if !isNewInstall {
		//TODO: It will update the TOTALS in the version table adding +1 just for the launches
	}
	
	return nil
}
```

Reasons for this approach
* Make clear and less code required
* It has high cohesion 
* It is following the good practices and respecting the layers.
* if you add the totals in the service in the struct you will be more susceptible to face concurrency issues since between the time to get the object and the value in this layer until the moment of persisting the data it could, in fact, be already changed. 

I hope that it help

pkg/web/apps/apps_service.go
camilamacedo86: @craicoverflow please check the suggestion https://github.com/aerogear/mobile-security-service/pull/76#discussion_r

pkg/web/apps/apps_psql_repository.go
camilamacedo86: Check my suggestion https://github.com/aerogear/mobile-security-service/pull/76#discussion_r which also solved this scenario.

pkg/web/apps/apps_repository.go
camilamacedo86: I know that it is possible to be done in the database. However, please check my suggestion https://github.com/aerogear/mobile-security-service/pull/76#discussion_r which also will solve this scenario.

pkg/web/apps/apps_service.go
camilamacedo86: @davidffrench the version of some app should not be inserted by the init call. If we do it we will create any version send from the SDK. Is it really makes sense? It is something that I do not remember we discussed about, however, shows more one action that will be trigged by the OCP. How a developer deploy a new client version in Aerogear? In RHMAP it is done in the build tab of the client app. c/c @craicoverflow

pkg/web/apps/apps_service.go
craicoverflow: Update the DeviceVersion property of the Device model.

camilamacedo86: IHMO it is not required. 
The device strtuct should have constructors. 
In this way you can have: `model.Device.New(deviceInfo, version)`  
which will be something as 

```go
func New(deviceInfo ..., version ..) modelDevice{
      return &device{
         params
       }
}
```

Also, after you have the device you can do things like `device.versionInfoN = version.InfoN`

Is it make sense?

camilamacedo86: I'd like to recommend you check how to create constructors. See it [here](https://golang.org/doc/effective_go.html#composite_literals) for example.

pkg/web/initclient/init_http_handler.go
davidffrench: It is currently specific to the init request. Not sure I would put this in the model. It is specific to the http layer in my opinion

pkg/web/apps/apps_service.go
camilamacedo86: Look for how to create constructors. see [here](https://golangbot.com/structs-instead-of-classes/) for example.

pkg/web/apps/apps_psql_repository.go
davidffrench: Hi @camilamacedo86 .

I really like the structure of the code, specifically calling out hasVersion and hasDevice, I think it is easier to read.

Unfortunately there is no way around having the `/init` create the version, this is the only time we will get to know a new version automatically without manual input from a user.

The sample code in your comment does not cover all scenarios either, I have updated the flow chart to account for all flows of the init endpoint. The DB creates/updates could esaily be inserts as well.

mss init flow

camilamacedo86: @davidffrench  I checked the chart added here. I could not fund one scenario that is not covered by the suggestion. Note that:

- if has a version and has deviceXVersion then it is the launch scenario 
- if has a version but has not deviceXVersion then   it is the new install
- if has NOT a version and NOT a deviceXVersion it is new install and a new version

What situation is not covered?

pkg/web/apps/apps_service.go
craicoverflow: @camilamacedo86 this is not a constructor function. This is updating a specific property on the device model based on a condition.

pkg/web/apps/apps_service.go
craicoverflow: Yes I moved this into a ctor. Thanks for the suggestion.

pkg/web/initclient/init_http_handler_test.go
camilamacedo86: Could you not do something as in order to make clear and easier: 

```go
deviceInvalidId = helpers.GetMockDevice()
deviceInvalidId.deviceId = "invalid" 

```

pkg/web/apps/apps_psql_repository.go
davidffrench: @camilamacedo86 I don't see the last scenario covered where the versionID needs to be updated for a device if the versionID does not match on the retrieved device.

pkg/web/initclient/init_http_handler_test.go
craicoverflow: I don't see the point. Each case has a different mock device, with different missing properties. We _could_ create a helper function for each of these test cases, but what is the point? We would be writing the same amount of code, except the unit test would now be dependent on an external package.

pkg/web/apps/apps_psql_repository.go
camilamacedo86: Hi @davidffrench yes it us true. Following the same with this last scenario. 

#### If has a version and has tracked a device using this version then it is the launch scenario 
**Actions**
* update the total launch in the version table

#### If has a version but has NOT tracked a device using this version then it is the new install
**Actions**
* Check if the device exist in the database with any other version if the SAME app
* If NOT exist then -> create the device 
* If exist then -> update the version number in the device 
* update the total launch and install in the version table
####  If has NOT the version then it is a new device install of a new version
**Actions**
* create the version 
* create the device 
* update the totals

c/c @craicoverflow

pkg/web/initclient/init_http_handler_test.go
craicoverflow: How do you suggest this can be done in the test table format?

```go
{
		{
			name: "A 400 Bad Request should be returned when request body is missing device ID",
			args: args{
				device: models.Device{
					ID:            "d1895cc1-28d7-4283-932d-8bcab9e4a469",
					VersionID:     "1a3fbe59-e69c-42c7-b9c0-d7eb7e65c073",
					AppID:         "com.aerogear.testapp",
					DeviceVersion: "1.1",
					Version:       "9.1.1",
					DeviceType:    "Android",
				},
			},
			wantStatusCode: 400,
		},
		{
			name: "A 400 Bad Request should be returned when request body is missing version",
			args: args{
				device: models.Device{
					ID:            "d1895cc1-28d7-4283-932d-8bcab9e4a469",
					DeviceID:      "1a3fbe59-e69c-42c7-b9c0-d7eb7e65c073",
					AppID:         "com.aerogear.testapp",
					DeviceVersion: "1.1",
					DeviceType:    "Android",
				},
			},
			wantStatusCode: 400,
		},
		{
			name: "A 400 Bad Request should be returned when request body is missing app ID",
			args: args{
				device: models.Device{
					ID:            "d1895cc1-28d7-4283-932d-8bcab9e4a469",
					VersionID:     "1a3fbe59-e69c-42c7-b9c0-d7eb7e65c073",
					DeviceVersion: "1.1",
					Version:       "9.1.1",
					DeviceType:    "Android",
				},
			},
			wantStatusCode: 400,
		},
		{
			name: "Expect init data to be returned when valid device is supplied",
			args: args{
				device: models.Device{
					ID:            "d1895cc1-28d7-4283-932d-8bcab9e4a469",
					VersionID:     "1a3fbe59-e69c-42c7-b9c0-d7eb7e65c073",
					DeviceID:      "db511711-95e8-4da2-8e76-1700465ae8ca",
					AppID:         "com.aerogear.testapp",
					DeviceVersion: "1.1",
					Version:       "9.1.1",
					DeviceType:    "Android",
				},
			},
			wantStatusCode: 200,
		},
```

camilamacedo86: You will put it one top as global value for the func and pass a device as arg.

pkg/web/apps/apps_service.go
camilamacedo86: I still not getting the need for this impl. What is the need and when it should be done? Why you need do this implementation? To solve what scenario?

pkg/web/initclient/init_http_handler_test.go
camilamacedo86: I added [here](https://github.com/aerogear/mobile-security-service/pull/76#discussion_r262887466) the detailed explanation and how I am suggesting you do it. 

The main reasons for you use the mock are:
* Make clear what is diff in each scenario, what changed and what will be checked
* Make easier the understanding and keep maintained. 
* Reduce  significantly the complexity of the func( it is not a good practice have fun with more than X lines which means with high complexity)
* Centralize the implementation which means that if the struct change we need change just the mock instead of understanding all these ones and do change across all tests. For example, we decided to add a property value for the versions then all tests need to be changed + mocks. If you follow the suggestion when you add the new property and value in the mock all will still work well as should be.

pkg/web/apps/apps_service.go
camilamacedo86: I am closing this one as resolved so.

pkg/web/apps/apps_service.go
craicoverflow: Because if the condition is met then an update takes place. See [here](https://github.com/aerogear/mobile-security-service/blob/3fa0f2a5693f3b79102015088846dd1d5a1678f9/pkg/web/apps/apps_service.go#L157-L158)

pkg/web/apps/apps_service_test.go
craicoverflow: @camilamacedo86 I cannot use the global mockRepository here. As you can see I am conditionally checking if the data from the test table matches that of the argument supplied in the mockRepository:

```go
GetAppByAppIDFunc: func(appID string) (*models.App, error) {
    if tt.fields.app == nil {
        return nil, models.ErrNotFound
    }

    return tt.fields.app, nil
},
```

The mock repositories you define are completely independent of the data from the test tables. They will not actually correctly mock the repository.

pkg/web/apps/apps_service.go
camilamacedo86: It still not making sense. 


Why you need a func updateDeviceDeviceVersion + the ifs and code implementations in the service when it is simple as follows.? 

> IF NOT exist a device with ANY version of the same app
> *  create the device
> ELSE
> * update the version number in the device

The service implementation made here is too unnecessary complex, as you check in the [suggestion](https://github.com/aerogear/mobile-security-service/pull/76#discussion_r262879241) which is a pseudo code it can be really more simple, clean, and with high cohesion, and more understandable. If get the text and replace it for the code you will have the service implemented. 

Note that the solution implemented here as I described above will definitely face concurrency issues and in this way will not attend the no-functional requirements of this system. 

c/c @davidffrench

pkg/web/apps/apps_http_handler_test.go
craicoverflow: > Hi, @craicoverflow I am afraid that it is not a good scenario for another task. It is the kind of thing that can be lost with time. IMHO, leave logs and outputs in the code is not a good idea and make easier we forget this and do not come back to fix it.

But this test case stopped working after a rebase so it is not really in the scope of this task/PR. The test is failing so we will not forget about it.

pkg/web/initclient/init_http_handler_test.go
```diff
@@ -0,0 +1,99 @@
...
+	tests := []struct {
+		name           string
+		fields         fields
+		args           args
+		wantStatusCode int
+	}{
+		{
+			name: "A 400 Bad Request should be returned when request body is missing device ID",
+			args: args{
+				device: helpers.getMockDeviceWithMissingDeviceID(),
```
camilamacedo86: Should be better you have just one mock and update it as suggested. 
It is better in this way than before, however, the code still not centralized. 

See how it will be with the suggestion.
**In the helper you will have just one mock for the device**
Then in the test func you can do as.

```go
func TestHTTPHandler_InitClientApp(t *testing.T) {
deviceWithMissingDeviceID:= helpers.GetMockDevice()
deviceWithMissingDeviceID.deviceID = ""

deviceWithMissingVersion:= helpers.GetMockDevice()
deviceWithMissingDeviceID.version = ""

{...}

     tests := []struct {
		name    string
		fields  fields
		device model.Device
		wantErr error
	}{
		{
			name: "A 400 Bad Request should be returned when request body is missing device ID",
			device: deviceWithMissingDeviceID
                         wantStatusCode: 400,
		},
		{
			name: "A 400 Bad Request should be returned when request body is missing version",
			device: deviceWithMissingVersion
                         wantStatusCode: 400,
		},
	}

{...}
```
Is it make sense? 
PS.: Anyway, it is better than before.

pkg/web/apps/apps_http_handler_test.go
camilamacedo86: But for we are able to merge it should pass in the CI. Make we can do a task just to implement the test but how can we ensure what was imp without the tests?

pkg/web/apps/apps_psql_repository.go
camilamacedo86: I will close this one because it is too big and also few requirements changed already. 
I will open a new conversation. 
c/c @davidffrench and @craicoverflow

pkg/models/device.go
```diff
@@ -1,11 +1,42 @@
...
+		DeviceID:      deviceID,
+		DeviceVersion: deviceVersion,
+		DeviceType:    deviceType,
+	}
+}
+
+// ValidateInitBody validates the properties of an init
+// request and returns an error if any of them are missing
+func (d *Device) ValidateInitBody() error {
+	if d.Version == "" || d.AppID == "" || d.DeviceID == "" {
```
camilamacedo86: It is not part of the model. It is related to the handler/router layer.

pkg/web/apps/apps_service.go
```diff
@@ -109,3 +111,90 @@ func (a *appsService) BindingAppByApp(appId, name string) error {
 	// if is deleted so just reactive the existent app
 	return a.repository.UnDeleteAppByAppID(app.AppID)
 }
+
+// // InitClientApp returns information about the current state of the app - its disabled status
+func (a *appsService) InitClientApp(deviceInfo *models.Device) (*models.Version, error) {
```
camilamacedo86: Hi @craicoverflow,

The best way that I found to help you with it here is doing the PR: https://github.com/aerogear/mobile-security-service/pull/88/files

NOTE: 
* Now, we also need to create or update the deleted_at of the app 
* has some checks that we missed in the previous discussions too.  
* see how the constructors were made and have been used. 

I am pinging @davidffrench in order to ask for he also checks the helping/suggestion PR in order and give his thoughts.

pkg/web/apps/apps_service_test.go
camilamacedo86: I am afraid that I could not clarifies it for you. Let's move forward we can do a refactor in another PR. CLosig it.

pkg/web/apps/apps_psql_repository.go
camilamacedo86: I got why it is here now 👍

pkg/web/apps/apps_service_test.go
camilamacedo86: I think that you get that very well :-) we can mark it as resolved.

pkg/web/initclient/init_http_handler_test.go
camilamacedo86: HI @craicoverflow it could be done as [here](https://github.com/aerogear/mobile-security-service/pull/90/files#diff-b910f1425e179a0e8b94ac700b838479R382)

pkg/models/device.go
craicoverflow: @camilamacedo86 I moved this based on your suggestion.

davidffrench: This should be in the http layer in my opinion. It is validating the request body of the init endpoint specifically.

craicoverflow:
> Hi @craicoverflow,
> 
> Unfortunately, IMHO we cannot move forward with this PR and merge it because of the following reasons. It is my review over it, however, I will leave it for @davidffrench to take the decision to approve or not.
> 
> * It has no tests to be executed in the first comment to check at least manually all scenarios implemented to ensure that they are working well expected. (E.g new install of an existent version in the client mobile, new launch and new install for a new version and check the totals of launches and installs)
> * The new [InitHandler](https://github.com/aerogear/mobile-security-service/pull/76/files#diff-335ad0864a7ce6d1010c42f6c0734300R107) implemented here shows go against the @davidffrench's definitions. All should be in the appHandler and has no need for a new Handler.
> * The changes performed here are not all covered by unit test either and it is not passing in the coveralls
> * The way that it is implemented will face concurrence issues
> * The implementation made is not following all good practices and a pattern which allow us to do the changes that indeed are already required in order to address all acceptance criteria and keep it maintained easily
> 
> The details of the changes suggested and reasons are here: #90 (PS.: 2 new comments were made there in order to explain better what was behind of the changes suggested in the service layer and how in the future we could move forward)

Hi @camilamacedo86 thank you for your feedback.

I moved the incrementation of `NumOfAppLaunches` to the database layer as you suggested, in order to deal with the concurrency issues you raised. I cannot see a reference to any other concurrency issues in https://github.com/aerogear/mobile-security-service/pull/90?

There was an agreement to keep the InitHandler. The reason being is that it follows the API URL pattern. So because the init endpoint is in a separate path to the `/apps` it makes sense to have a separate handler, because it is handling a whole new different path.

cc @davidffrench

camilamacedo86:
Hi @craicoverflow,

> I moved the incrementation of NumOfAppLaunches to the database layer as you suggested, in order to deal with the concurrency issues you raised. I cannot see a reference to any other concurrency issues in #90?

The explanation over concurrence issues is [here](https://github.com/aerogear/mobile-security-service/pull/90/files#r263252783) and in the service implementation. You will see how it was implemented to +1 when it is a new install and when it is a new launch. Just check the scenarios and the repo method called. 

However, I also note that the new install has been not saved in the database as a total too and it has been made by the `select count(device.id) where version = {}`.  So, after a fast re-check in your PR shows that it is solved for now. I forgot it and it the reason we are using current installs instead of total installs. :-)

pkg/web/apps/apps_psql_repository.go
```diff
@@ -186,24 +311,6 @@ func (a *appsPostgreSQLRepository) CreateApp(id, appId, name string) error {
 	return nil
 }
 
-func (a *appsPostgreSQLRepository) GetAppByAppID(appID string) (*models.App, error) {
```
camilamacedo86: Hi @craicoverflow, 

Before change or remove a method is required check if it has been or not used in the project. In this case, it is used in the service binding and cannot be removed or changed in order to filter and return just the active ones. There is required to get the app by app id when it is disabled too. 

To return apps which are just active the project has already a method implemented which is GetActiveAppID the is also a good idea follow the same nomenclature. I recommend you create a method called GetActiveAppByAppID and keep this one.

Also, note that when a dev will use a method GetObjectByID is not intuitive at all expected that it has been filtered by any other thing that is not the appID. It is important to highlight that indeed the init will use fetch the disabled apps too since it will do the binding steps in the future. Is it make sense?

craicoverflow: > Hi @craicoverflow,
> 
> In this case, it is used in the service [binding](https://github.com/aerogear/mobile-security-service/blob/a0aae35c13836f3250a1190707797e7072d0ca6f/pkg/web/apps/apps_service.go#L97) and cannot be removed or changed in order to filter and return just the active ones. There is required to get the app by app id when it is disabled too.
> 
> To return apps which are just active the project has already a method implemented which is GetActiveAppID then, in my point of view, is also a good idea follow the same nomenclature. I recommend you create a method called GetActiveAppByAppID and keep this one.
> 
> Also, note that when a dev will use a method GetObjectByID is not intuitive at all expected that it will be filtering by any other thing that is not the appID, do not you agree?. It is important to highlight that indeed the /init will fetch the disabled apps too in the neext steps when it will be creating/updating apps, I mean doing the binding. Is it make sense? Could you please do this fix before merge?

Hi @camilamacedo86 

I will do the change in order to remain consistent but I disagree with the sentiment. 

They are **deleted** not **inactive**. It should be assumed that an endpoint will never return deleted rows from the database, it should not need to be specified in the function name.

And if we are going the way of having appropriate naming then it would not be `Get**Active**AppByAppID` because at no stage are the apps ever active or inactive.

camilamacedo86: Hi @craicoverflow,

Following some considerations:
* The name delete was a not good choice since the app is NOT DELETED at all as its versions and device.
* If an App is updated as "deleted_at" it should not return in the search, however, it can be undeleted again and according to all definitions, its versions and devices should come back to appear. So, it is more active and inactive app. 
* The endpoint is one layer and the repository methods are another. If you are implementing a service in the REST you will never suppose that a method with the name GetAppByID will return just the active ones or where the deleted_at is not null or not "soft deleted" as you can choose the name.

craicoverflow: > Hi @craicoverflow,
> 
> Following some considerations:
> 
> * The name delete was IHMO a not good choice at all since the app is NOT DELETED as its versions and device and never will be.
> * If an App has an updated as "deleted_at" it should not return in the search, however, it can be undeleted again and according to all definitions, its versions and devices should come back to appear too. So, according to the requirements, it is more like to disable an app then delete, do not you agree?
> * The endpoint is one layer and the repository methods are another. If you are implementing a service in the REST you will never assume that a method with the name GetAppByID will return just the active ones or where the deleted_at is not null or not "soft deleted" as you can prefer the nomenclature.
> * Before remove or change the implementation of any method in any project is mandatory to check if it has or not been used. By the IDE you can check where something is or not used easily.  Is it make sense? We cannot change the implementation of a method that has been used for other services.

Hi @camilamacedo86

To help me understand your considerations, where in the requirements does it say that apps can or will be undeleted?

craicoverflow: Hi @camilamacedo86 I have updated the function name to be consistent with the rest of the server.

camilamacedo86: Hi @craicoverflow,

Following the answers. 

1. To help me understand your considerations, where in the requirements does it say that apps can or will be undeleted?

* Actions/Task: The unbinding and binding services. 
* You can check the comments in the JIRA to know how it was defined. 

2. According to your considerations, do you also think that the endpoint URL should be modified to include this information?

NO. The endpoint should not be changed. It is required just to keep the method in the repo which has been used for the binding service.

pkg/web/apps/apps_repository.go
```diff
@@ -13,6 +13,11 @@ type Repository interface {
 	DisableAllAppVersionsByAppID(appID string, message string) error
 	DeleteAppByAppID(appId string) error
 	CreateApp(id, appId, name string) error
-	GetAppByAppID(appID string) (*models.App, error)
```
camilamacedo86: Here, it is missing came back to the project this method. Also, the binding service should still be using this one and not the new one GetActiveAppByAppID
@craicoverflow @davidffrench

pkg/web/apps/apps_service_test.go
```diff
@@ -33,7 +35,7 @@ var (
 		CreateAppFunc: func(id string, appId string, name string) error {
 			return nil
 		},
-		GetAppByAppIDFunc: func(appID string) (*models.App, error) {
+		GetActiveAppByAppIDFunc: func(appID string) (*models.App, error) {
```
camilamacedo86: Shows that you are removing the mocks as when you should just add the new ones. 

Followinng example.
```go
GetAppByAppIDFunc: func(appID string) (*models.App, error) {
			return nil, models.ErrNotFound
		},
```

Please, compare the mock tests with the master branch and ensure that you just added your new ones.

craicoverflow: Hi @camilamacedo86 good spot. This was done from the previous change when we thought we only needed one function. Changed now.

pkg/web/apps/apps_repository.go
craicoverflow: Fixed @camilamacedo86

PR #81, opened 2019-03-04
feat(binding) - Implement binding app service and repository (AEROGEAR-8725)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8725

## What

- Implement service and its test which will be called when an app is binding
- Implement the repository to create a new app when the binding is made and its tests.

## Why
It will be used when the service is binding in the app. 

## Verification Steps
1. Check if all unit tests finish with success
2. Check if the queries added are working with success manually in the database.
a) use the seed to insert data in the database
b) use the query in the psql method CreateApp as the following example

```sql
INSERT INTO app (id, app_id, app_name) VALUES ('be2da1f5-a9c4-4305-84bc-80da683fbc36', 'com.aerogear.test', 'test');
```
c) check the query in the psql method UnDeleteAppByAppID as following example.  

```sql
		UPDATE app
		SET deleted_at=NULL
		WHERE app_id='com.aerogear.test'
```

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

screenshot 2019-03-04 at 17 59 02

screenshot 2019-03-04 at 18 00 04

### DIFF
--- a/pkg/web/apps/apps_service.go
+++ b/pkg/web/apps/apps_service.go
@@ -1,6 +1,7 @@
 package apps
 
 import (
+	"github.com/aerogear/mobile-security-service/pkg/helpers"
 	"github.com/aerogear/mobile-security-service/pkg/models"
 )
 
@@ -12,6 +13,7 @@ type (
 		UpdateAppVersions(versions []models.Version) error
 		DisableAllAppVersionsByAppID(id string, message string) error
 		UnbindingAppByAppID(appID string) error
+		BindingAppByApp(appId, name string) error
 	}
 
 	appsService struct {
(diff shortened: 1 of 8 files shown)


pkg/web/apps/apps_service.go
```diff
@@ -79,3 +81,17 @@ func (a *appsService) DisableAllAppVersionsByAppID(id string, message string) er
...
+func (a *appsService) BindingAppByApp(appId, name string) error {
+
+	app, err := a.repository.GetAppByAppID(appId)
+
+	if err != nil && err == models.ErrNotFound {
+		id := helpers.GetUUID()
+		return a.repository.CreateApp(id, appId, name)
+	} else if err != nil {
+		return err
+	} else {
```
davidffrench: It would be better to split this logic into two seperate functions in my opinion to deal with the `create` and `undelete`.

camilamacedo86: Done 👍 thank you for the review.

PR #83, opened 2019-03-05
fix(ci): CI not running fmt issues discovered by pre-commit hook - (AEROGEAR-8750)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8750

## What
Fix script to return an error regards the code format

## Why
The CI should return an error when the code is not formatted

## How
Changing the command and using validation for that. 

## Verification Steps
- Since this PR has not the fix https://github.com/aerogear/mobile-security-service/pull/82 then it should fail in the CI. 

Also, is possible to check if after format it will work as expected.  
- Cherry pick the  https://github.com/aerogear/mobile-security-service/pull/82 to get all fmt
- Then check that the CI will return success after that
OR
*Locally:*
- Execute the `git commit` and see the same msg
- Run `make fmt`
- After fmt the files check that the git commit will finished with success. 

## Checklist:

## Progress

- [] Finished task
- [X] TODO - Testing

## Additional Notes

 
screenshot 2019-03-05 at 09 48 45

screenshot 2019-03-05 at 09 53 04

### DIFF
--- a/scripts/code-check.sh
+++ b/scripts/code-check.sh
@@ -2,9 +2,16 @@
 
 # Run some pre commit checks on the Go source code. Prevent the commit if any errors are found
 
-# Format the Go code
+# Check if the Go code is formatted
 check_code_format (){
-    go fmt $(go list ./... | grep -v /vendor/)
+    gofmt -l . | grep -v vendor/ >/dev/null 2>&1
+    if [ $? -ne 1 ]; then
+       printf "\nErrors found in your code, please use 'make fmt' to format your code."
+       exit 1
+    else
+       exit 0
+    fi
+
 }
 
 # Check all files for errors


scripts/code-check.sh
```diff
@@ -2,9 +2,16 @@
 
 # Run some pre commit checks on the Go source code. Prevent the commit if any errors are found
 
-# Format the Go code
+# Check if the Go code is formatted
 check_code_format (){
-    go fmt $(go list ./... | grep -v /vendor/)
+    gofmt -l . | grep -v vendor/ >/dev/null 2>&1
+    if [ $? -ne 1 ]; then
```
davidffrench: Will this approach still format the code for the user, so that they can just stage the files and commit?

camilamacedo86: It will check if the user would like to format the code them is required to use the command "make fmt" for example. Would you like that both happens?

laurafitzgerald: I think it's better to ask the user to run the command. I had some problem with files getting changed locally when I was trying to interactively rebase to update some commits and I kept getting the additional files added there without me confirming it. The files that were being synced had no relation to my changes as they were in the go server and I was working on the ui. I didn't make sense to include them in that pr.

camilamacedo86: I agree with you @laurafitzgerald sametimes the magic make the things more complicated than easier and make more sense shows the error and not fix without the user knows what and why.

davidffrench:
Great fix, thank you @camilamacedo86

PR #84, opened 2019-03-05
fix(deleted-at): add deleted_at to sql for GetActiveAppById (AEROGEAR-8742)

austincunningham:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8742

## What
Renames GetAppsById to GetActiveAppsById
Added `WHERE deleted_at IS NULL` to GetActiveAppsById
Added tests

## Why
It is expected behavior that deleted apps don't return from the endpoint

## How
`WHERE deleted_at IS NULL` to GetActiveAppsById

## Verification

- Run the server and check the following url
```bash
go run cmd/mobile-security-service/main.go
```
Seed the database with
```sql
INSERT INTO app
			(id, app_id, app_name, deleted_at)
		VALUES 
			('1b9e7a5f-af7c-4055-b488-72f2b5f72266', 'com.aerogear.foobar', 'Foobar', NULL),
			('ae2da1f5-a9c4-4305-84bc-80da683fbc36', 'com.test.app1', 'App One', '2019-02-18 14:36:35'),
			('0890506c-3dd1-43ad-8a09-21a4111a65a6', 'com.aerogear.testapp', 'Test App', NULL);

		INSERT INTO version
			(id, version, app_id, disabled, disabled_message, num_of_app_launches)
		VALUES 
			('f6fe70a3-8c99-429c-8c77-a2efa7d0b458', '1', 'com.aerogear.testapp', FALSE, '', 5000),
    	('9bc87235-6bcb-40ab-993c-8722d86e2201', '1.1', 'com.aerogear.testapp', TRUE, 'Please contact an administrator', 1000),
			('def3c38b-5765-4041-a8e1-b2b60d58bece', '1', 'com.test.app1', FALSE, '', 10000);
				
		INSERT INTO device
			(id, version_id, app_id, device_id, device_type, device_version)
		VALUES 
			('d19feeb4-fb21-44e8-9990-473bf97a0a3f', 'f6fe70a3-8c99-429c-8c77-a2efa7d0b458', 'com.aerogear.testapp', 'a742f8b7-5e2f-43f3-a3c8-073da858420f', 'iOS', '10.2'),
			('00cb8957-db04-4ab6-8fd8-14b9fc516dbd', '9bc87235-6bcb-40ab-993c-8722d86e2201', 'com.aerogear.testapp', 'd1895cc1-28d7-4283-932d-8bcab9e4a461', 'Android', '3.2'),
			('e3b43b01-167b-48ef-8ff4-caf2e6613dee', '9bc87235-6bcb-40ab-993c-8722d86e2201', 'com.aerogear.testapp', 'feee7f81-0e33-4548-abbb-17a681c12f3b', 'Android', '4.1'),
			('ab411c3e-29f8-4e70-9ddc-8bafbba3fc4c', 'def3c38b-5765-4041-a8e1-b2b60d58bece', 'com.test.app1', '94da9833-093e-4f4c-9a93-b11600ce46b7', 'iOS', '2.0'),
			('a42a128a-dfb6-435c-8653-8f66ab3a5a1c', 'def3c38b-5765-4041-a8e1-b2b60d58bece', 'com.test.app1', '94132b0c-d7b1-4419-bcce-fc6760c59e3a', 'Android', '4.1');
```

- Soft deleted apps requests like 
http://localhost:3000/api/apps/ae2da1f5-a9c4-4305-84bc-80da683fbc36
will return 
```js
{
  "message: "Your requested Item is not found",
  "statusCode": 404
}
```

## Checklist:

## Progress

- [ ] Finished task
- [ ] TODO

## Additional Notes

PS.: Add images and/or .gifs to illustrate what was changed if this pull request modifies the appearance/output of something presented to the users.

### DIFF
--- a/pkg/web/apps/apps_service.go
+++ b/pkg/web/apps/apps_service.go
@@ -9,7 +9,7 @@ type (
 	// Service defines the interface methods to be used
 	Service interface {
 		GetApps() (*[]models.App, error)
-		GetAppByID(ID string) (*models.App, error)
+		GetActiveAppByID(ID string) (*models.App, error)
 		UpdateAppVersions(versions []models.Version) error
 		DisableAllAppVersionsByAppID(id string, message string) error
 		UnbindingAppByAppID(appID string) error
@@ -40,10 +40,10 @@ func (a *appsService) GetApps() (*[]models.App, error) {
 	return apps, nil
 }
 
-// GetAppByID retrieves app by id from the repository
-func (a *appsService) GetAppByID(id string) (*models.App, error) {
+// GetActiveAppByID retrieves app by id from the repository
+func (a *appsService) GetActiveAppByID(id string) (*models.App, error) {
 
-	app, err := a.repository.GetAppByID(id)
+	app, err := a.repository.GetActiveAppByID(id)
 
 	if err != nil {
 		return nil, err
@@ -74,7 +74,7 @@ func (a *appsService) UpdateAppVersions(versions []models.Version) error {
 func (a *appsService) DisableAllAppVersionsByAppID(id string, message string) error {
 
 	// get the app id to send it to the re
-	app, err := a.repository.GetAppByID(id)
+	app, err := a.repository.GetActiveAppByID(id)
 
 	if err != nil {
 		return err
--- a/pkg/web/router/router.go
+++ b/pkg/web/router/router.go
@@ -50,7 +50,7 @@ func SetAppRoutes(r *echo.Group, appsHandler apps.HTTPHandler) {
 	// Retrieve all information for a single app including all child information
 	// ---
 	// summary: Get app by id
-	// operationId: getAppById
+	// operationId: GetActiveAppByID
 	// produces:
 	// - application/json
 	// parameters:
@@ -68,8 +68,7 @@ func SetAppRoutes(r *echo.Group, appsHandler apps.HTTPHandler) {
 	//     description: Invalid id supplied
 	//   404:
 	//     description: App not found
-	r.GET("/apps/:id", appsHandler.GetAppByID) // TODO: Implement correctly the call of the method passing the parameters
-
+	r.GET("/apps/:id", appsHandler.GetActiveAppByID)
 	// swagger:operation PUT /apps/:id/versions Version
 	//
 	// Update all versions informed of an app using the app id, including updating version information
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -158,7 +158,7 @@ paths:
   /apps/{id}:
     get:
       description: Retrieve all information for a single app including all child information
-      operationId: getAppById
+      operationId: GetActiveAppByID
       parameters:
       - description: The id for the app that needs to be fetched.
         in: path
--- a/pkg/web/apps/apps_http_handler.go
+++ b/pkg/web/apps/apps_http_handler.go
@@ -13,7 +13,7 @@ import (
 type (
 	HTTPHandler interface {
 		GetApps(c echo.Context) error
-		GetAppByID(c echo.Context) error
+		GetActiveAppByID(c echo.Context) error
 		UpdateAppVersions(c echo.Context) error
 		DisableAllAppVersionsByAppID(c echo.Context) error
 	}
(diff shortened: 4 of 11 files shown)


pkg/web/router/router.go
```diff
@@ -68,7 +68,7 @@ func SetAppRoutes(r *echo.Group, appsHandler apps.HTTPHandler) {
 	//     description: Invalid id supplied
 	//   404:
 	//     description: App not found
```
camilamacedo86: IMHO will be go add in the descriptions that this router and apps just return active apps

pkg/web/apps/apps_service.go
```diff
@@ -74,7 +74,7 @@ func (a *appsService) UpdateAppVersions(versions []models.Version) error {
 func (a *appsService) DisableAllAppVersionsByAppID(id string, message string) error {
 
 	// get the app id to send it to the re
-	app, err := a.repository.GetAppByID(id)
+	app, err := a.repository.GetActiveAppByID(id)
```
camilamacedo86: Hi @davidffrench just to confirm that are you ok with this point? 
After this change the /apps/{id}/versions/disable endpoint will just work for the apps which are active. I mean it will not work with those ones which have the `deleted_at != NULL`. 

@austincunningham and @davidffrench , it is not a blocker to merge this PR at all, I just would like to highlight this for @davidffrench. Because if we would like to keep the endpoint `disable all versions` working for all apps and not just the active ones then we can just to create a small task to use here a method which GetAppByID without a filter by active ones. Anyway, of this change shows not go against the current requirements as well.

pkg/web/router/router.go
```diff
@@ -68,7 +68,7 @@ func SetAppRoutes(r *echo.Group, appsHandler apps.HTTPHandler) {
...
-	r.GET("/apps/:id", appsHandler.GetAppByID) // TODO: Implement correctly the call of the method passing the parameters
+	r.GET("/apps/:id", appsHandler.GetActiveAppByID) // TODO: Implement correctly the call of the method passing the parameters
```
camilamacedo86: Also, could you remove the TODO comments? This endpoint is already done and then it should no longer be here.e

pkg/web/apps/apps_service.go
austincunningham: makes sense

pkg/web/router/router.go
austincunningham: :+1:

pkg/web/router/router.go
austincunningham: so `Active App not found` suggests there may be an active app when there may not be, I think `App not found` should be enough for a customer to see on the endpoint

pkg/web/apps/apps_service.go
davidffrench: Hey @camilamacedo86 . The `/apps/{id}/versions/disable` should only work for apps that are not soft deleted in my opinion. Soft deleted apps from a users point of view are deleted.

pkg/web/router/router.go
austincunningham: lol removing the comment reduced the code coverage

pkg/web/router/router.go
camilamacedo86: @davidffrench,

Should we really do it or instead of all this change just add in the service layer to return in the response one error telling that the app searched is inactive? Wdyt? @austincunningham

davidffrench: Are there currently multiple errors returned?

It should just be the one error of `app not found`.

I might be missing the ask here.

camilamacedo86: My suggestion was an app return app not found or app disable when it is disabled or the app.
instead of the not found for both scenarios.

camilamacedo86:
@davidffrench and @austincunningham 

1. The changes here shows fine ( I think we can move forward)
2. I made a suggestion regards a second option ( but I am fine with as it is now )
3. Just the coveralls that decreased a little because the validator in the router has no tests. See [here](https://coveralls.io/builds/21996198/source?filename=pkg/web/router/router.go)

So, I am approving this one and I will let you folks decide.

PR #85, opened 2019-03-06
feat: navigate to app on row click

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8552

## What
Clicking on the Table now routes the user to app detailed view passing the id in the route

## Why
See Motivation

## How
- Adding the data to the apps state value
- Using the data to pick up the db id of the app using the app name
- pushing a new router to the react router history

## Verification Steps
 
_Functionality_
1. Run `make serve`
2. Ensure there are records in the db
3. Navigate to /
4. Click on one of the rows in the table 
5. Verify the id passed in the route matches the app clicked on using data in the database.

_Tests_
1. Run `npm test`
2. expect all tests to pass

## Checklist:

## Progress

- [x] Functionaity
- [x] Tests

### DIFF
--- a/ui/src/containers/AppsTableContainer.js
+++ b/ui/src/containers/AppsTableContainer.js
@@ -22,7 +22,6 @@ export class AppsTableContainer extends React.Component {
     const id = app[0].id;
     const path = '/app/' + id;
     this.props.history.push(path);
-    // return <Redirect to={location}/>;
   }
   onSort (_event, index) {
     this.props.reverseAppsTableSort(index);
--- a/ui/src/containers/AppsTableContainer.js
+++ b/ui/src/containers/AppsTableContainer.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import AppsTable from '../components/AppsTable';
 import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { getApps, reverseAppsTableSort } from '../actions/actions-ui';
 
@@ -15,8 +16,13 @@ export class AppsTableContainer extends React.Component {
     this.props.getApps();
   }
   onRowClick (event, rowId, props) {
-    console.log('On row click called');
-    this.setState({ redirect: true });
+    var app = this.props.apps.data.filter(app => {
+      return app.appName === rowId[0];
+    });
+    const id = app[0].id;
+    const path = '/app/' + id;
+    this.props.history.push(path);
+    // return <Redirect to={location}/>;
   }
   onSort (_event, index) {
     this.props.reverseAppsTableSort(index);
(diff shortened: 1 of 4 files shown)


austincunningham:
Can't verify, maybe missing some steps in the verification

ui/src/reducers/index.js
```diff
@@ -37,16 +37,18 @@ export default (state = initialState, action) => {
       var fetchedApps = [];
       action.result.forEach(app => {
         var temp = [];
-        temp[0] = app.appId;
```
camilamacedo86: Will the appID appears in another place? If not,  wdyt over keep both infos?

ui/src/containers/AppsTableContainer.js
```diff
@@ -15,8 +16,13 @@ export class AppsTableContainer extends React.Component {
     this.props.getApps();
   }
   onRowClick (event, rowId, props) {
-    console.log('On row click called');
-    this.setState({ redirect: true });
+    var app = this.props.apps.data.filter(app => {
+      return app.appName === rowId[0];
```
camilamacedo86: Is possible more than one app has the same name? Will, it works well expected in this scenario?

laurafitzgerald: @camilamacedo86 you can't create an app with a display name that has the same as a previous one in mdc.
screenshot 2019-03-06 at 12 51 47
I agree this isn't an elegant solution but we are tied somewhat with how the Table is being built in patternfly

ui/src/reducers/index.js
laurafitzgerald: Not sure what the question is here. This is managing the state of the apps page so it would only appear once on that page. We'll probably have a different object for the state of the detailed apps ui which will get populated on the get request for the app/{id}

laurafitzgerald:
> Can't verify, maybe missing some steps in the verification

Hi @austincunningham what step did you face problems with. I added an additional step that there should be data in the db.

ui/src/reducers/index.js
laurafitzgerald: Ah I understand the question now. This array is what's used to render the table and we use the appName in the table and not the appId as we discussed in the chat.

laurafitzgerald: In the lines below you can see that I'm storing the full data object returned by the request. See `data: action.result`

ui/src/containers/AppsTableContainer.js
camilamacedo86: Usually, the solution is to hidden the info required which should not be shown to the user as follows. 

```css
.hidden {
display: none;
}

<div class="hidden"> content </div>
```

ui/src/reducers/index.js
camilamacedo86: IMHO both info is important to the users since the appName is the name given for them to create the app and the app_id is the data that the mobile device via SDK should send in order to make it works.

laurafitzgerald:
Resolving the conversations with @camilamacedo86 we are happy with the searching by appname approach given that the appname will be unique. This is managed in MDC. Camilla will leave a comment in a trello card regarding the other one as we may want to show the app id as well in the table. Will resolved that separately if required.

laurafitzgerald:
Based on the above @camilamacedo86 @austincunningham are you guys okay to approve?

camilamacedo86:
The code for shows great and @austincunningham confirmed that he verified it and all is working. So, I am approving.

PR #88, opened 2019-03-06
collab Enda

camilamacedo86:
## Motivation
https://github.com/aerogear/mobile-security-service/pull/76

## What
* Service, Repo and model imps as an idea/suggestion to help/collab with the task.

## Why
* The requirements changed and it is not a trivial or easy task at all. 
* It was the easy way that I could found to try to show some suggestions and help and collab with

## How
- Get the master branch
- Get the repo layer changes of the PR https://github.com/aerogear/mobile-security-service/pull/76
- Doing a few improvements in the repo layer
- Doing the init call endpoint in the service implementation in order to try to cover all scenarios
- Creating the mock files to ensure that all are declared ok in the service and repo layer
- Checked that all test still working with success

### DIFF
--- a/pkg/web/apps/apps_psql_repository.go
+++ b/pkg/web/apps/apps_psql_repository.go
@@ -218,3 +218,126 @@ func (a *appsPostgreSQLRepository) UnDeleteAppByAppID(appId string) error {
 
 	return nil
 }
+
+// GetVersionByAppIDAndVersion gets a version by its app ID and version number
+func (a *appsPostgreSQLRepository) GetVersionByAppIDAndVersion(appID string, versionNumber string) (*models.Version, error) {
+	version := models.Version{}
+
+	sqlStatement := `
+	SELECT v.id,v.version,v.app_id, v.disabled, v.disabled_message, v.num_of_app_launches, v.last_launched_at
+	FROM version as v
+	WHERE v.app_id = $1 AND v.version = $2;`
+
+	err := a.db.QueryRow(sqlStatement, appID, versionNumber).Scan(&version.ID, &version.Version, &version.AppID, &version.Disabled, &version.DisabledMessage, &version.NumOfAppLaunches, &version.LastLaunchedAt)
+
+	if err != nil {
+		log.Error(err)
+		if err == sql.ErrNoRows {
+			return nil, models.ErrNotFound
+		}
+		return nil, models.ErrInternalServerError
+	}
+
+	return &version, nil
+}
+
+// GetDeviceByDeviceIDAndAppID returns a device by its device ID and app ID
+func (a *appsPostgreSQLRepository) GetDeviceByDeviceIDAndAppID(deviceID string, appID string) (*models.Device, error) {
+	device := models.Device{}
+
+	sqlStatement := `
+		SELECT d.id, d.version_id, d.app_id, d.device_id, d.device_type, d.device_version
+		FROM device as d
+		WHERE d.device_id = $1 AND d.app_id = $2;`
+
+	if err := a.db.QueryRow(sqlStatement, deviceID, appID).
+		Scan(&device.ID, &device.VersionID, &device.AppID, &device.DeviceID, &device.DeviceType, &device.DeviceVersion); err != nil {
+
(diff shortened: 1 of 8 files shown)


pkg/web/apps/apps_repository.go
```diff
@@ -14,5 +14,12 @@ type Repository interface {
 	DeleteAppByAppID(appId string) error
 	CreateApp(id, appId, name string) error
 	GetAppByAppID(appID string) (*models.App, error)
+	GetActiveAppByAppID(appID string) (*models.App, error)
```
camilamacedo86: We need thee GetAppByID and the GetActiveAppByAppID.
We also need re-check this points @austincunningham

pkg/web/apps/apps_psql_repository.go
```diff
@@ -218,3 +218,126 @@ func (a *appsPostgreSQLRepository) UnDeleteAppByAppID(appId string) error {
...
+		log.Error(err)
+		return err
+	}
+
+	return nil
+}
+
+func (a *appsPostgreSQLRepository) IncrementVersionTotals(versionID string, isNewInstall bool) error {
+	//TODO
+	// Example to try to avoid concurrence issues by delegating to the DB the increase count and update
```
camilamacedo86: Here is the reality something as :

```
	// UPDATE versions
	// SET
	// total = total + 1
	// WHERE
	// id={$1}
```

pkg/web/apps/apps_service.go
```diff
@@ -109,3 +110,179 @@ func (a *appsService) BindingAppByApp(appId, name string) error {
...
+	if hasAnyOtherVersion {
+		if err := a.repository.UpdateDeviceVersion(deviceDiffVersion, version); err != nil {
+			return err
+		}
+	} else {
+		if err := a.repository.CreateNewDevice(device); err != nil {
+			return err
+		}
+	}
+	return nil
```
camilamacedo86: This part if you prefer by the DB as you made and follows.

```go
// InsertDeviceOrUpdateVersionID creates a new device row in the device table
func (a *appsPostgreSQLRepository) InsertDeviceOrUpdateVersionID(device *models.Device) error {
	sqlStatement := `
		INSERT INTO device(id,version_id,app_id,device_id,device_type,device_version)
		VALUES($1, $2, $3, $4, $5, $6)
		ON CONFLICT (id)
		DO UPDATE
		SET version_id = $2, device_version = $6`

	_, err := a.db.Exec(sqlStatement, device.ID, device.VersionID, device.AppID, device.DeviceID, device.DeviceType, device.DeviceVersion)

	if err != nil {
		log.Error(err)
		return models.ErrInternalServerError
	}

	return nil
}
```

pkg/web/apps/apps_service.go
```diff
@@ -109,3 +110,179 @@ func (a *appsService) BindingAppByApp(appId, name string) error {
...
+			return nil, err
+		}
+		a.repository.IncrementVersionTotals(version.ID, true)
+	}
+
+	// ** New version and new install Scenario **
+	// If has NOT the version then it is a new device install of a new version
+	if !hasVersion && !hasDeviceVersion {
+		version := models.NewVersion(sdkInfo, app)
+		version.ID = helpers.GetUUID()
```
camilamacedo86: This 2 lines should go inside of the func as follows. 

```go
func (a *appsService) handleNewVersionAndInstall(sdkInfo, deviceDiffVersion *models.Device, app *models.App, hasAnyOtherVersion bool) error {
	version := models.NewVersion(sdkInfo, app)
	version.ID = helpers.GetUUID()
	
	if err := a.repository.CreateNewVersion(version); err != nil {
		return err
	}
	if err := a.createOrUpdateDevice(hasAnyOtherVersion, deviceDiffVersion, version, sdkInfo, app); err != nil {
		return err
	}
	a.repository.IncrementVersionTotals(version.ID, true)
	return nil
}


```

camilamacedo86:
IGNORE. I made the above comments and add it here : https://github.com/aerogear/mobile-security-service/pull/89

PR #94, opened 2019-03-11
fix(style): adding style to the apps detail table (AEROGEAR-8766)

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8766
 
## What
Styling
- The table
- I also improved the blank slate somewhat

## Why
It's nice when things look nice

## How
css magic

## Verification Steps

1. Run `make serve`
2. UI should appear as below

Screenshot 2019-03-11 at 13 18 21

## Checklist:

## Additional Notes

Pushing this as a WIP as I'm basing it on the styling as outlined in the story flat but I did make some additions that weren't asked for so happy to remove those if necessary.

### DIFF
--- a/ui/src/actions/__tests__/actions-ui.test.js
+++ b/ui/src/actions/__tests__/actions-ui.test.js
@@ -1,15 +1,18 @@
 import * as actions from '../actions-ui';
-import { REVERSE_SORT } from '../types';
+import { APPS_SORT } from '../types';
+import { SortByDirection } from '@patternfly/react-table';
 
 describe('actions', () => {
   it('should reverse the app Table Sort', () => {
     const index = 0;
+    const direction = SortByDirection.asc;
     const expectedAction = {
-      type: REVERSE_SORT,
+      type: APPS_SORT,
       payload: {
-        index: index
+        index: index,
+        direction: direction
       }
     };
-    expect(actions.reverseAppsTableSort(index)).toEqual(expectedAction);
+    expect(actions.appsTableSort(index, direction)).toEqual(expectedAction);
   });
 });
--- a/ui/src/components/AppsTable.js
+++ b/ui/src/components/AppsTable.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Table, TableHeader, TableBody } from '@patternfly/react-table';
 import { withRouter } from 'react-router-dom';
+import './Table.css';
 
 export const AppsTable = ({ columns, rows, sortBy, onSort, onRowClick }) => (
   <Table aria-label="Mobile Apps" sortBy={sortBy} onSort={onSort} cells={columns} rows={rows}>
--- a/ui/src/components/LandingPage.css
+++ b/ui/src/components/LandingPage.css
@@ -1,0 +1,5 @@
+.title {
+    margin-left: 5%;
+    margin-top: 2%;
+    margin-bottom: 2%
+}
--- a/ui/src/components/LandingPage.js
+++ b/ui/src/components/LandingPage.js
@@ -1,6 +1,8 @@
 import React from 'react';
+import { Title } from '@patternfly/react-core';
 import Header from './common/Header';
 import AppsTableContainer from '../containers/AppsTableContainer';
+import './LandingPage.css';
 
 class LandingPage extends React.Component {
   constructor () {
@@ -12,6 +14,7 @@ class LandingPage extends React.Component {
     return (
       <>
         <Header />
+        <Title className="title" size="3xl">Mobile Apps</Title>
         <AppsTableContainer />
       </>
     );
--- a/ui/src/components/Table.css
+++ b/ui/src/components/Table.css
@@ -1,0 +1,8 @@
+.apps-table {
+    margin-left: 5%;
+    margin-right: 5%;
+}
+
+button.pf-c-button {
+    text-transform: uppercase;
+}
--- a/ui/src/components/common/Header.js
+++ b/ui/src/components/common/Header.js
@@ -10,6 +10,7 @@ import {
   ToolbarGroup,
   ToolbarItem
 } from '@patternfly/react-core';
+import { withRouter } from 'react-router-dom';
 import { UserIcon } from '@patternfly/react-icons';
 import accessibleStyles from '@patternfly/patternfly/utilities/Accessibility/accessibility.css';
 import { css } from '@patternfly/react-styles';
@@ -19,8 +20,10 @@ import { connect } from 'react-redux';
 import config from '../../config/config';
 import { toggleHeaderDropdown } from '../../actions/actions-ui';
 
-class Header extends React.Component {
-  onTitleClick = () => {};
+export class Header extends React.Component {
+  onTitleClick = () => {
+    this.props.history.push('/');
+  };
 
   onUserDropdownToggle = () => {
     this.props.toggleHeaderDropdown();
@@ -35,10 +38,6 @@ class Header extends React.Component {
   };
 
   render () {
-    const logoProps = {
-      onClick: () => this.onTitleClick()
-    };
-
     const toolbar = (
       <Toolbar>
         <ToolbarGroup className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnLg)}>
@@ -61,7 +60,7 @@ class Header extends React.Component {
       </Toolbar>
     );
 
-    const Header = <PageHeader logo={config.app.name.toUpperCase()} logoProps={logoProps} toolbar={toolbar} />;
+    const Header = <PageHeader logo={config.app.name.toUpperCase()} onClick={this.onTitleClick} toolbar={toolbar} />;
 
     return (
       <div className="mssHeader">
@@ -83,4 +82,4 @@ function mapStateToProps (state) {
   };
 }
 
-export default connect(mapStateToProps, { toggleHeaderDropdown })(Header);
+export default withRouter(connect(mapStateToProps, { toggleHeaderDropdown })(Header));
--- a/ui/src/containers/AppsTableContainer.js
+++ b/ui/src/containers/AppsTableContainer.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import { withRouter } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { getApps, reverseAppsTableSort } from '../actions/actions-ui';
+import './TableContainer.css';
 
 export class AppsTableContainer extends React.Component {
   constructor (props) {
@@ -39,7 +40,7 @@ export class AppsTableContainer extends React.Component {
     if (this.props.isAppsRequestFailed) {
       return (
         <div className="no-apps">
-          <p>Unable to fetch any apps</p>
+          <p>Unable to fetch any apps :/</p>
         </div>
       );
     }
--- a/ui/src/containers/AppsTableContainer.js
+++ b/ui/src/containers/AppsTableContainer.js
@@ -3,7 +3,7 @@ import AppsTable from '../components/AppsTable';
 import PropTypes from 'prop-types';
 import { withRouter } from 'react-router-dom';
 import { connect } from 'react-redux';
-import { getApps, reverseAppsTableSort } from '../actions/actions-ui';
+import { getApps, appsTableSort } from '../actions/actions-ui';
 import './TableContainer.css';
 
 export class AppsTableContainer extends React.Component {
@@ -16,7 +16,7 @@ export class AppsTableContainer extends React.Component {
   componentWillMount () {
     this.props.getApps();
   }
-  onRowClick (event, rowId, props) {
+  onRowClick (_event, rowId) {
     var app = this.props.apps.data.filter(app => {
       return app.appName === rowId[0];
     });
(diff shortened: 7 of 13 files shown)


ui/src/components/LandingPage.css
```diff
@@ -0,0 +1,11 @@
+.title {
+    text-align: center;
```
damienomurchu:
```suggestion
    text-align: center;
    text-align: left;
```

ui/src/components/Table.css
```diff
@@ -0,0 +1,12 @@
+.apps-table {
```
damienomurchu: @laurafitzgerald can we increase the size of the table head font?

laurafitzgerald: @damienomurchu sure. Should I make it lower case then?

ui/src/components/LandingPage.css
laurafitzgerald: I didn't like it aligned to the left. But it is aligned left in the flat so I will do that.

ui/src/components/LandingPage.js
```diff
@@ -12,6 +14,8 @@ class LandingPage extends React.Component {
     return (
       <>
         <Header />
+        <Title className="title" size="4xl">Mobile Apps</Title>
```
damienomurchu:
```suggestion
        <Title className="title" size="2xl">Mobile Apps</Title>

```

damienomurchu: `4x1` seems larger than Title font in flat

ui/src/components/common/Header.js
```diff
@@ -19,7 +20,9 @@ import { connect } from 'react-redux';
 import { toggleHeaderDropdown } from '../../actions/actions-ui';
 
 class Header extends React.Component {
-  onTitleClick = () => {};
+  onTitleClick () {
+    this.props.history.push('/');
```
damienomurchu: Are we using this navigation history/ is it wired into React already?

laurafitzgerald: It uses Router. the history objection tracks the navigation across all components. It works really well. You can check it by navigating between the home page and the two apps and then using the back button.

ui/src/components/LandingPage.js
laurafitzgerald: I tried out the 2xl. Went with 3xl in the end as it seemed to look the nicer of the 3 options.

laurafitzgerald:
@damienomurchu @craicoverflow I've pushed some new changes which add some fixes other than the styling. There's a method and action name change and also some fixes on the navigation and some state setting which was a little broken too. I also updated tests alongside those fixes. Would you mind reviewing again when you get a chance.

ui/src/components/Table.css
laurafitzgerald: I added some styling to the table header and I think it makes it stand out now without increasing the font.

damienomurchu:
@laurafitzgerald will review shortly. One q - added css files don't match the component file name they're being used with, is this deliberate?

ui/src/containers/AppsTableContainer.js
```diff
@@ -15,22 +16,22 @@ export class AppsTableContainer extends React.Component {
   componentWillMount () {
     this.props.getApps();
   }
-  onRowClick (event, rowId, props) {
+  onRowClick (_event, rowId) {
```
damienomurchu: Is `_event` param required (as it's not being used) ?

ui/src/containers/AppsTableContainer.js
```diff
@@ -15,22 +16,22 @@ export class AppsTableContainer extends React.Component {
...
     var app = this.props.apps.data.filter(app => {
       return app.appName === rowId[0];
     });
     const id = app[0].id;
     const path = '/app/' + id;
     this.props.history.push(path);
   }
-  onSort (_event, index) {
-    this.props.reverseAppsTableSort(index);
+  onSort (_event, index, direction) {
```
damienomurchu: Is `_event` param required (as it's not being used)

ui/src/actions/__tests__/actions-ui.test.js
```diff
@@ -1,15 +1,18 @@
 import * as actions from '../actions-ui';
-import { REVERSE_SORT } from '../types';
+import { APPS_SORT } from '../types';
```
damienomurchu: I wonder is this descriptive enough to distinguish it from the sort we will be doing on the the AppDetails table, or do you have a suggestion on what the sort action for AppDetails should be named so we keep them consistent?

ui/src/containers/AppsTableContainer.js
craicoverflow: It is the first parameter in a default function, it cannot be removed.

ui/src/components/AppsTable.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Table, TableHeader, TableBody } from '@patternfly/react-table';
 import { withRouter } from 'react-router-dom';
+import './Table.css';
```
damienomurchu: Should this be `AppsTable.css` ?

ui/src/containers/AppsTableContainer.js
craicoverflow: It is the first parameter in a default function, it cannot be removed.

ui/src/actions/__tests__/actions-ui.test.js
laurafitzgerald: I would suggest APP_SORT for the app table

ui/src/components/AppsTable.js
laurafitzgerald: I was thinking we could use it for both tables as we probably want them to be consistent which is why I left the name more generic.

ui/src/components/common/Header.js
```diff
@@ -18,8 +19,17 @@ import PropTypes from 'prop-types';
...
-  onTitleClick = () => {};
+export class Header extends React.Component {
+  constructor (props) {
+    super(props);
+    this.onTitleClick = this.onTitleClick.bind(this);
+    this.onUserDropdownToggle = this.onUserDropdownToggle.bind(this);
+    this.onUserDropdownSelect = this.onUserDropdownSelect.bind(this);
+    this.onLogoutUser = this.onLogoutUser.bind(this);
+  }
+  onTitleClick () {
```
damienomurchu:
```suggestion
  onTitleClick = () => {
```

damienomurchu: Converting to arrow function as suggested will remove need to bind `this` to the method in the constructor

ui/src/components/common/Header.js
```diff
@@ -18,8 +19,17 @@ import PropTypes from 'prop-types';
...
 
-class Header extends React.Component {
-  onTitleClick = () => {};
+export class Header extends React.Component {
+  constructor (props) {
+    super(props);
+    this.onTitleClick = this.onTitleClick.bind(this);
+    this.onUserDropdownToggle = this.onUserDropdownToggle.bind(this);
+    this.onUserDropdownSelect = this.onUserDropdownSelect.bind(this);
+    this.onLogoutUser = this.onLogoutUser.bind(this);
```
damienomurchu: The `this` binding being done here is not necessary as `this` is being bound below by using arrow functions to define the handlers. 

If all `bind` calls are removed, this constructor can be eliminated too

ui/src/components/common/Header.js
```diff
@@ -60,7 +66,7 @@ class Header extends React.Component {
       </Toolbar>
     );
 
-    const Header = <PageHeader logo="MOBILE SECURITY SERVICE" logoProps={logoProps} toolbar={toolbar} />;
+    const Header = <PageHeader logo="MOBILE SECURITY SERVICE" onClick={this.onTitleClick} toolbar={toolbar} />;
```
damienomurchu: Is `onClick` supported by the PageHeader component? (Documentation seems to indicate it isn't - http://patternfly-react.surge.sh/patternfly-4/components/page)

laurafitzgerald: I believe it's just a property. I was being passed as a property in the previous implementation also.

ui/src/reducers/index.js
```diff
@@ -25,19 +25,19 @@ const initialState = {
 
 export default (state = initialState, action) => {
   switch (action.type) {
-    case REVERSE_SORT:
-      const reversedOrder = state.sortBy.direction === SortByDirection.asc ? SortByDirection.desc : SortByDirection.asc;
+    case APPS_SORT:
       const index = action.payload.index;
-      const sortedRows = state.apps.rows.sort((a, b) => (a[index] < b[index] ? -1 : a[index] > b[index] ? 1 : 0));
-      const sortedApps = reversedOrder === SortByDirection.asc ? sortedRows : sortedRows.reverse();
+      var sortedRows = state.apps.rows.sort((a, b) => (a[index] < b[index] ? -1 : a[index] > b[index] ? 1 : 0));
```
damienomurchu: Change from var to let here

ui/src/components/common/Header.js
damienomurchu: Can you share where it was being passed previously? 

I haven't ran the code to check if the onClick call works here - do you know if it does?

laurafitzgerald: It's in the same file. I removed a few lines above this.

craicoverflow:
Header does not stay full width when viewed on a smaller screen. This was also happening before this PR (switched back to master to check) but now the gap has grown significantly so something here may be affecting it.

Screenshot from 2019-03-12 09-46-53

ui/src/components/common/Header.js
damienomurchu: Can you confirm the `onClick()` still works, as before it was being passed in as part of the object to the logoProps attribute to PageHeader. PageHeader doesn't seem to accept an `onClick` property[1], and looks like an onClick handler can only be passed in via logoProps to PageHeader

[1]: https://github.com/patternfly/patternfly-react/blob/0b637feaf467a860ee69263bb2c477a3d0090c22/packages/patternfly-4/react-core/src/components/Page/PageHeader.js#L8

laurafitzgerald: I can confirm that it's working.

ui/src/components/LandingPage.css
```diff
@@ -0,0 +1,5 @@
+.title {
+    padding-top: .5em;
+    margin-bottom: .5em;
+    margin-left: 2em;
```
damienomurchu:
```suggestion
    margin-left: 1em;
```
With `2em` the title looks too indented - `1em` aligns the header text, title text, and table text, wdyt?

ui/src/components/common/Header.js
damienomurchu: Me too, resolving this now

damienomurchu:
@laurafitzgerald, PR looks good, just one comment left re left padding of Title

ui/src/components/LandingPage.css
laurafitzgerald: When I use 1em I see this: 

Screenshot 2019-03-12 at 11 55 44

Could you post a screenshot also?

damienomurchu: Screenshot from 2019-03-12 11-59-53
Screenshot from 2019-03-12 12-00-06

damienomurchu: With `2em`:
Screenshot from 2019-03-12 12-02-53
Screenshot from 2019-03-12 12-03-09

damienomurchu: Screenshot from 2019-03-12 12-04-10

laurafitzgerald:
@craicoverflow the responsive header issue should be sorted with the new commit. Would you mind verifying that when you get a chance.

ui/src/components/LandingPage.css
laurafitzgerald: I'm using Chrome: Version 72.0.3626.109. Strange that I have exactly the same and it looks very different. Would the version of Chrome make that much difference?

laurafitzgerald: @damienomurchu pushed a new change which should resolve this. Would you mind verifying?

ui/src/components/LandingPage.css
```diff
@@ -1,5 +1,5 @@
 .title {
-    padding-top: .5em;
-    margin-bottom: .5em;
-    margin-left: 2em;
+    margin-left: 5%;
```
damienomurchu: @laurafitzgerald is there a reason for the change in units used for the styling?

laurafitzgerald: yes, i believe we'll have less inconsistency. When I set both the table and the Header to have 2em they didn't line up. When using percentage they will line up.

laurafitzgerald: @damienomurchu does it fix the discrepancy on your end?

damienomurchu:
@laurafitzgerald reviewed. Fix for overflowing table is probably the best we can hope for with the present implementation of the table in PF4 which does not appear to be as responsive as it should be compared to patternfly-core components. 

Just had 1 question left in comments. Think you addressed everything else

ui/src/components/LandingPage.css
damienomurchu: Understand. Ideally I think we would look at using `rem` (which is used by PF, and the spacing for the components specified usually in `rem` at the bottom of each component page). If we need to get this merged in due to time constraints we can go with `%` but I think `rem` might be better (but we can refactor in future).

damienomurchu: Left comment above

PR #95, opened 2019-03-11
feat(app-detailed-view): Make a GET to the API to retrieve an app by its ID (AEROGEAR-8466)

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8466

## What

Added logic to make a GET request to the API and fetch a single app by its ID.

## Why

It is required so that we the UI can fetch live, dynamic content from the server.

## How

Added new function to `DataService.js` and updated Redux store to handle new state changes.

## Verification Steps
 
1. Make sure the API is running.
2. On the landing page, click one of the app rows to bring you to the app detailed page.
3. Open the developer console and monitor the output from the redux-logger.
4. If you see a log with action type `APP_SUCCESS` then the UI has successfully retrieved the app information from the server.
5. Navigate to the app details view by clicking on one of the apps on the landing page.
6. The table of deployed versions should render and should be sortable on all columns.

## Checklist:

## Progress

- [x] Finished task


ui/src/actions/types.js
```diff
@@ -1,6 +1,10 @@
 export const REVERSE_SORT = 'REVERSE_SORT';
 export const GET_APPS = 'GET_APPS';
```
laurafitzgerald: @craicoverflow am I right in saying we don't use either GET_APPS or GET_APP_BY_ID anywhere?

ui/src/reducers/__tests__/index.test.js
```diff
@@ -1,38 +1,41 @@
...
+      { title: 'Last Launched', transforms: [sortable] },
+      { title: 'Disable on Startup', transforms: [sortable] },
+      { title: 'Custom Disable Message', transforms: [sortable] }
     ],
     isAppsRequestFailed: false,
     currentUser: 'currentUser',
-    isUserDropdownOpen: false
+    isUserDropdownOpen: false,
+    app: {
+      deployedVersions: { rows: [], data: {} }
```
laurafitzgerald: @craicoverflow there seems to already be a `deployedVersions` array inside the data object. Could we remove it from here so that the data object contains the following:


Screenshot 2019-03-14 at 17 19 33

As it stands it's nested as below:

Screenshot 2019-03-14 at 17 20 08

ui/src/reducers/index.js
```diff
@@ -99,19 +104,43 @@ export default (state = initialState, action) => {
...
+    case APPS_FAILURE:
+      return {
+        ...state,
+        isAppsRequestFailed: true
+      };
+    case APP_REQUEST:
+      return {
+        ...state
+      };
+    case APP_SUCCESS:
```
laurafitzgerald: I think you included this because we do the translation for the apps request in the reducer also. But because there is a React Component (the checkbox) required in the appversions table @damienomurchu did the translation in the component rather than in the reducer. We discussed moving the apps rows translation to the component again, but it's not necessary that we do that. For this case, as we're already doing it in the component we don't need to do it here.

laurafitzgerald: In that case we probably don't even need to store the rows value in the state as we're creating that in the component and can use the json data to build it there.

craicoverflow: Ah, I originally had it like that but changed it to keep the component changes minimal. I kept the component code mostly the same:

https://github.com/aerogear/mobile-security-service/pull/95/files#diff-2aac7a8dcccd74e488dbea5673dfc9c2R14

I will change it back.

craicoverflow: I think we might need to store the row values in the state for sorting.

ui/src/containers/AppVersionTableContainer.js
```diff
@@ -55,31 +52,29 @@ export class AppTableContainer extends React.Component {
...
 
 function mapStateToProps (state) {
   return {
-    app: state.appDetailRows,
-    sortBy: state.appDetailsSortDirection,
-    columns: state.appDetailColumns,
-    isAppsRequestFailed: state.isAppsRequestFailed
+    sortBy: state.appVersionsSortDirection,
+    columns: state.appVersionsColumns,
+    appVersions: state.app.deployedVersions.rows
```
damienomurchu: Should we include also appVersions under `AppVersionTableContainer.propTypes` above?

craicoverflow: Would that fail when appVersions is an empty array or is that valid?

ui/src/actions/types.js
craicoverflow: You are right.

ui/src/containers/AppVersionTableContainer.js
damienomurchu: The propType check shouldn't - my understanding is it just checks for the presence of that prop and triggers a warning if it's not

ui/src/reducers/index.js
laurafitzgerald: @craicoverflow yep, you're 100% right. Resolving this one and there's no action required.

damienomurchu: > I think we might need to store the row values in the state for sorting.

I think you're right. We only get the re-rendering of the table component with redux when the state mapped in `mapPropsToState` in the component changes. 

With the `AppVersionTableContainer` we need to modify the row data to create the checkbox, and we shouldn't be storing React components (like Checkbox'es) in the store, so I think we need logic in both places; in the reducer to store the row data in the state, and in the component to re-render the row data when it changes (& generate the checkboxes)

ui/src/containers/AppVersionTableContainer.js
craicoverflow: Updated.

ui/src/reducers/__tests__/index.test.js
laurafitzgerald: Marking as resolved. thanks for the update

laurafitzgerald:
I saw one small edge case where ordering on a column where all values are the same (e.g. `Last Launched` doesn't work as expected but let's resolve that in a seperate pr so that we can unblock the others. @craicoverflow would you mind creating a JIRA for that.

craicoverflow:
> I saw one small edge case where ordering on a column where all values are the same (e.g. `Last Launched` doesn't work as expected but let's resolve that in a seperate pr so that we can unblock the others. @craicoverflow would you mind creating a JIRA for that.

Yeah I see what you mean. Will do

PR #101, opened 2019-03-12
fix(swagger): Updated swagger doc to reflect real endpoint URL (AEROG…

craicoverflow:
…EAR-XXXX)

## Motivation

Endpoint URL was wrong in Swagger definition file.

## What

Updated the endpoint from `/apps/:id/versions/batch-disable` to `/apps/:id/versions/disable`.

## Why

Doc was outdated.

## Verification Steps
 
Verify that [this line](https://github.com/aerogear/mobile-security-service/blob/42b4e26663de578ebbb7f0fa0bde7fdc3c51b231/api/swagger.yaml#L133) says `/apps/:id/versions/disable:`

## Checklist:

## Progress

- [x] Finished task


api/swagger.yaml
```diff
@@ -130,7 +130,7 @@ paths:
         "404":
           description: App not found
       summary: Update 1 or more versions of an app
-  /apps/:id/versions/batch-disable:
+  /apps/:id/versions/disable:
```
camilamacedo86: It is missing update the tests which call the endpoint

craicoverflow: Probably a separate PR as a new JIRA as this is just fixing the Swagger doc.

craicoverflow: I will do it here for the sake of time though.

PR #102, opened 2019-03-12
Fix docker image to install UI 

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8820

## What
Fix docker image to install UI and check it running when a new-app on OCP be created by it

## Why
When the project be installed in OCP the UI should works

## How
By adding in the Dockerfile the required steps to install and run the UI

## Verification Steps
 
1. Create a project in OCP
2. Create a Postgres from the catalogue with the default values of this project set up in 
3. Run `oc new-app https://github.com/camilamacedo86/mobile-security-service.git#FIX-UI --strategy=docker -e PGHOST=postgresql --name=app`
4. When the deploy finish create the router by `oc expose svc/app`
5. Click on in the router created and check if the UI appears

## Checklist:

## Progress

- [X] Finished task
- [] TODO 

## Additional Notes

Screenshot 2019-03-13 at 00 33 10

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,20 @@
 FROM openshift/origin-release:golang-1.10 as builder
-
 WORKDIR /go/src/github.com/aerogear/mobile-security-service
 COPY . .
 RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh \
 && dep ensure \
 && go build -o mobile-security-service ./cmd/mobile-security-service/main.go
 
+FROM node:10 as ui
+WORKDIR /go/src/github.com/aerogear/mobile-security-service
+COPY . .
+RUN cd ui/ && npm install --production && npm run build
+
 FROM centos:7
-
 COPY --from=builder /go/src/github.com/aerogear/mobile-security-service /usr/bin/
-
-EXPOSE 3000
+COPY --from=ui /go/src/github.com/aerogear/mobile-security-service/ui/build /opt/mobile-security-service/ui/build/
 
 USER 1001
-
+ENV STATIC_FILES_DIR /opt/mobile-security-service/ui/build
+EXPOSE 3000
 ENTRYPOINT ["/usr/bin/mobile-security-service"]


Dockerfile
```diff
@@ -2,9 +2,12 @@ FROM openshift/origin-release:golang-1.10 as builder
 
 WORKDIR /go/src/github.com/aerogear/mobile-security-service
 COPY . .
+RUN curl --silent --location https://rpm.nodesource.com/setup_10.x | bash -
+RUN yum -y install nodejs
 RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh \
 && dep ensure \
-&& go build -o mobile-security-service ./cmd/mobile-security-service/main.go
+&& go build -o mobile-security-service ./cmd/mobile-security-service/main.go \
+&& export STATIC_FILES_DIR=$(CURDIR)/ui/build; ./mobile-security-service -kubeconfig ~/.kube/config
```
camilamacedo86: Folks, it is not working. Any idea how we can solve it?

davidffrench: @camilamacedo86 Our Dockerfile should be the same as MDC, at least the parts we used similar to them.

https://github.com/aerogear/mobile-developer-console/blob/master/Dockerfile#L8

camilamacedo86: Thank you it was helpful.

craicoverflow:
I notice the reason you are implementing these changes is so that you can run `oc new-app` pointing to the Dockerfile in the GitHub repo. It would be better to run `oc new-app` pointing to our prebuilt Docker image hosted on quay.io or Docker Hub, like so:

```
oc new-app google/golang-hello
```

davidffrench:
> This is supposed to be a Dockerfile for a production environment but now we are installing things that are only required for development such as NPM, React-Scripts. I don't believe this stuff should be happening here.
> 
> cc @davidffrench @grdryn

I would be curious to get Gers opinion on this. I would have thought the changes were to allow the UI to work and be served from Go but this requires it to be built.

craicoverflow:
> > This is supposed to be a Dockerfile for a production environment but now we are installing things that are only required for development such as NPM, React-Scripts. I don't believe this stuff should be happening here.
> > cc @davidffrench @grdryn
> 
> I would be curious to get Gers opinion on this. I would have thought the changes were to allow the UI to work and be served from Go but this requires it to be built.

See my other comment https://github.com/aerogear/mobile-security-service/pull/102#issuecomment-472398608 on why I think this is being done and why I think it is not needed. Some of these changes are probably required, such as:

```Dockerfile
ENV STATIC_FILES_DIR /opt/mobile-security-service/ui/build
```

and 

```Dockerfile
COPY --from=builder /go/src/github.com/aerogear/mobile-security-service/ui/build /opt/mobile-security-service/ui/build/
```

But most of the rest should be already done before pushing to Quay.io/Docker Hub.

camilamacedo86:
Hi @craicoverflow 

> This is supposed to be a Dockerfile for a production environment but now we are installing things that are only required for development such as NPM, React-Scripts. I don't believe this stuff should be happening here.
> cc @davidffrench @grdryn
Comment:  https://github.com/aerogear/mobile-security-service/pull/102#issuecomment-472398608

The image in the Quay.io/Docker Hub is the Dockerfile, so do the command passing the image in the registry and/or the repository when this image is it is actually the same.  

> This bypasses our hosted Docker image. We should be creating the app in OpenShift from our hosted image on Quay.io with a stable release tag.

- The goal of this PR has an image that will install the Rest Service and UI.
- I understand that the creation of image was already implemented since I can find docker push command in the CI and any need/improvement/fix regards to it is not part of this PR. 

@grdryn could you help us. Do you see something that is wrong in this Dockerfile? Have you any objection?

c/c @davidffrench

grdryn:
When we'll get around to creating an offline Dockerfile.rhel7, we won't be able to build the UI as part of the image build (that will need to be built separately and copied in (see [here](https://github.com/aerogear/mobile-developer-console/blob/master/Dockerfile.rhel7#L7) for example).

I don't see a huge problem with building the UI in your upstream image here though. What I would strongly suggest if you're going this route though, is to use a separate builder image for the UI build, rather than installing Node.js into the Go builder image _every time_.

More like this for example:

```Dockerfile
FROM openshift/origin-release:golang-1.10 as builder
// Do the Go build stuff as normal

FROM registry.centos.org/centos/nodejs-8-centos7:latest as ui
// Do your Node.js/npm stuff

FROM centos:7
COPY --from=builder /go/src/github.com/aerogear/mobile-security-service /usr/bin/
COPY --from=ui /wherever/it/is/ /opt/mobile-security-service/ui/build/

//etc
```

camilamacedo86:
HI @grdryn and @craicoverflow and @davidffrench.

I will try to improve it with the @grdryn suggestion and ping you to approve it.

craicoverflow:
> Hi @craicoverflow
> 
> > This is supposed to be a Dockerfile for a production environment but now we are installing things that are only required for development such as NPM, React-Scripts. I don't believe this stuff should be happening here.
> > cc @davidffrench @grdryn
> > Comment:  [#102 (comment)](https://github.com/aerogear/mobile-security-service/pull/102#issuecomment-472398608)
> 
> The idea of using a docker image is have a container with all that is required to build, to install and configure some app are already there and running when you use it without the need to perform all this steps manually. To do that in this project, the UI is using NPM and React and it is not just required for development only.
> 
> * The only change that I can see here is using `npm install --production` instead of `npm install`, however, the NPM is required. +1 Done.
> * The react is required for production and the reason of install react-scripts is because it has been used to build the UI. `  "build": "react-scripts build",`
> 
> > This bypasses our hosted Docker image. We should be creating the app in OpenShift from our hosted image on Quay.io with a stable release tag.
> 
> * The goal of this PR is just build an image that will install the Rest Service and UI for we use in the operator.
> * Just to clarifies, the image in the Quay.io/Docker Hub is acctualy the Dockerfile published there with an tag. In this way, use the ocp command to create the pod passil the URL of this repo and branch or the image published in the registry will have the same result if both are using the same image/file.
> 
> **IMPORTANT: Any requirement regards publish and tag it in a registry is not a scope of this PR. I understand that project has tasks/JIRAs for that already**
> 
> @grdryn could you help us. Do you see something that is wrong in this Dockerfile? Have you any objection?
> 
> c/c @davidffrench

I'm pretty sure that the Docker Hub/Quay.io repo contains the built image - not the Dockerfile.

camilamacedo86:
hi @folks, 

It is working and updated with @grdryn recommendations. Please, could we move forward and merge it now?

davidffrench:
Great work @camilamacedo86

PR #72, opened 2019-02-25
Add the endpoints to update 1 or many versions of the app or disable all +  Replacing all asserts for testing in the service tests layer + more tests + remove asserts

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8356
https://issues.jboss.org/browse/AEROGEAR-8611
https://issues.jboss.org/browse/AEROGEAR-8612

## What
- Disable all versions of an app. 
- Update 1 or many versions of the app 
- Replacing all asserts for testing in the service tests layer
- Improve and increase the tests in the layers
- Remove the `AppPostgreSQLRepository` and `PostgreSQLRepositor` interfaces from pkg/web/apps/apps_psql_repository.go since they are not needed. The `NewPostgreSQLRepository` func uses the generic `Repository` interface.
- Remove the unuseful mock from apps_psql_repository since it is not used. 

## Why
- To attend the UI/project actions requirements
- To attend project definitions

## How
- adding endpoints
- adding tests
- adding service and repository layer implementations

## Verification Steps
- Verify if it passed in all tests
- Verify that the % of tests was increased
- Verify that the project is no longer using asserts
- The unit-test are covering the points, however, please feel free to test it manually as follows. 

#### Setup project and database to do the tests.
- Following the steps described [here](https://github.com/aerogear/mobile-security-service#131-by-using-the-docker-image-of-this-project) in the README
- Then after starting the server, you can create the mock database
- Use can use the [seed] file to do it. See [here](https://github.com/aerogear/mobile-security-service/blob/master/pkg/helpers/db_seed.go)

#### Check endpoint to update all versions by POSTMAN
- Call the endpoint GET `apps` in order to get a valid id
- Call the endpoint POST `/apps/:id/versions/disable` with a JSON with the version struct or empty as the following examples.

```json
{"disabledMessage": "text"}
```

screenshot 2019-02-28 at 19 29 33
screenshot 2019-02-28 at 19 29 11

- Check that all versions of the app with ID were updated with the message informed and with the disable as true. 
- The endpoint should return HTTP code 200

#### Check endpoint to update 1 or N versions by POSTMAN
- Call the endpoint GET `apps` in order to get a valid id
- Call the endpoint GET `apps/:id` to get the version struct
- Call the endpoint PUT `/apps/:id/versions` with a JSON with the []versions ( list of versions ).

```json
[
    {
        "id": "9bc87235-6bcb-40ab-993c-8722d86e2201",
        "disabled": true,
[...]

PR #103, opened 2019-03-13
feat(update-versions) : Add check to validate the app id of the versions sent

camilamacedo86:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8822

## What
Add a check to validate if the versions are from the app sent. 

## Verification Steps
1. Check the CI.
2. Following the manual test:
a) Call the endpoint by sending versions which are NOT associated with the app informed (Should return an error)
b) Call the endpoint by sending versions which are associated with the app informed (Should work)

PS.: Check here how to call the endpoint: https://github.com/aerogear/mobile-security-service/pull/72

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/web/apps/apps_http_handler.go
+++ b/pkg/web/apps/apps_http_handler.go
@@ -87,7 +87,7 @@ func (a *httpHandler) UpdateAppVersions(c echo.Context) error {
 	}
 
 	// Call service
-	errUpdate := a.Service.UpdateAppVersions(versions)
+	errUpdate := a.Service.UpdateAppVersions(id, versions)
 	if errUpdate != nil {
 		return httperrors.GetHTTPResponseFromErr(c, errUpdate)
 	}
--- a/pkg/web/apps/apps_http_handler.go
+++ b/pkg/web/apps/apps_http_handler.go
@@ -92,7 +92,7 @@ func (a *httpHandler) UpdateAppVersions(c echo.Context) error {
 		return httperrors.GetHTTPResponseFromErr(c, errUpdate)
 	}
 
-	return c.JSON(http.StatusOK, "")
+	return c.NoContent(http.StatusNoContent)
 }
 
 //UpdateApp returns a app updated with the ID in JSON format from the AppService
--- a/pkg/web/apps/apps_service.go
+++ b/pkg/web/apps/apps_service.go
@@ -43,7 +43,7 @@ func (a *appsService) GetApps() (*[]models.App, error) {
 	return apps, nil
 }
 
-// GetActiveAppByID retrieves app by id from the repository
+// GetActiveAppByID retrieves app by id from the repository where the deleted_at is NULL
 func (a *appsService) GetActiveAppByID(id string) (*models.App, error) {
 
 	app, err := a.repository.GetActiveAppByID(id)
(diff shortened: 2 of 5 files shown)


pkg/web/apps/apps_service.go
```diff
@@ -63,14 +63,25 @@ func (a *appsService) GetActiveAppByID(id string) (*models.App, error) {
...
+
+	app, err := a.repository.GetActiveAppByID(id)
 
-	// Check for errors and return the appropriate error to the handler
 	if err != nil {
 		return err
 	}
 
+	for i := 0; i < len(versions); i++ {
+		if versions[i].AppID != app.AppID {
```
craicoverflow: What if 2/3 version app IDs are correct? Should we still update those?

camilamacedo86: No. I don't think so. If it has any data inconsistent than the error should be returned. 
In this case, if any version informed in the list is not associated with the id of the app informed an error as Bad Request ( invalid id supplied ) will be returned. Is it make sense? @davidffrench do you agree with it?

davidffrench: This type of request should never happen from our UI, so it is sufficient to send a Bad Request on the first incorrect id in my opinion.

camilamacedo86: OK. The log was added. Could you please verify this one for we are able to move forward and merge it?

camilamacedo86:
@craicoverflow you can use the image if you prefer. See here https://cloud.docker.com/repository/docker/cmacedo/mobile-security-service/general

pkg/web/apps/apps_service.go
camilamacedo86: done.

pkg/web/apps/apps_service.go
```diff
@@ -63,14 +64,26 @@ func (a *appsService) GetActiveAppByID(id string) (*models.App, error) {
 }
 
 // GetApps retrieves the list of apps from the repository
```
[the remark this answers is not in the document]
camilamacedo86: changed bisides has no relations with.

pkg/web/apps/apps_service_test.go
```diff
@@ -208,39 +208,51 @@ func Test_appsService_DisableAllAppVersionsByAppID(t *testing.T) {
 }
 
 func Test_appsService_UpdateAppVersions(t *testing.T) {
+	version := helpers.GetMockAppVersionList()[0]
```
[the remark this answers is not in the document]
camilamacedo86: Yep. But in this case, I would like to use the first item of this list. I don't think it need to be changed now.

pkg/web/apps/apps_http_handler.go
```diff
@@ -92,7 +92,7 @@ func (a *httpHandler) UpdateAppVersions(c echo.Context) error {
 		return httperrors.GetHTTPResponseFromErr(c, errUpdate)
 	}
 
-	return c.JSON(http.StatusOK, "")
```
craicoverflow: Should be

```go
return c.NoContent(http.StatusNoContent)
```

otherwise you are returning an empty string (which is content).

camilamacedo86: It is really something that will no make any diff for us. However, as you wish it is changed.

camilamacedo86:
Hi @craicoverflow,

All changes requested by you related or not to the PR are made. Please, could you check and let us know if we can move forward here?

PR #91, opened 2019-03-07
feat(UI): Adds App Detail view (AEROGEAR-8468)

damienomurchu:
## Motivation
Implement an App Details view (AEROGEAR-8468)

## What
Implement an App Details view

## Why
View details of an app

## How
Implement components, wire up to redux, and add tests

## Verification Steps
```shell
# ensure all tests pass
npm test

# start ui dev server
npm start

Verify:
- table renders with sample data
- table sorts correctly
- no error output when verifying changes
- no warning output when verifying changes
```

## Checklist:

- [x] Code has been tested locally by PR requester
- [ ] Changes have been successfully verified by another team member 

## Progress

- [x] Finished task

PR #105, opened 2019-03-13
Aerogear 8470

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8470

## What
Breadcrumb and Buttons

## How
Using PF4 toolbar and some css hacks

## Verification Steps
 
1. Run 'make serve'
2. The page should look as below

Screenshot 2019-03-13 at 18 05 59
 

3. Click on the Home breadcrumb
4. The page should navigate to the Landing Page
5. Navigation history should also work

## Checklist:

## Progress

- [x] Functionality
- [x] Rebase on [Table pr](https://github.com/aerogear/mobile-security-service/pull/91) once merged
- [x] Rebase on [get request pr](https://github.com/aerogear/mobile-security-service/pull/95) once merged
- [ ] Tests


ui/src/components/AppDetailToolbar.js
```diff
@@ -0,0 +1,64 @@
...
+      redirect: false
+    });
+  }
+
+  onDisableButton = () => {
+    console.log('Disable button clicked');
+  }
+
+  onClick = () => {
+    this.setState({
```
davidffrench: This will need to open up the modal if the current form state is dirty.

ui/src/components/AppDetailToolbar.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
+import { Breadcrumb, BreadcrumbItem, Toolbar, ToolbarGroup, Button, ToolbarItem } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+import './AppDetailedToolbar.css';
+
+class AppDetailToolbar extends React.Component {
+  componentWillMount () {
```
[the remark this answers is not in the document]
laurafitzgerald: @davidffrench after the discussion yesterday I thought we weren't going to implement that everywhere yet?

ui/src/components/AppDetailToolbar.js
laurafitzgerald: @davidffrench yes, this is a temporary usage until we have the the redux state implementation complete. @damienomurchu will be making some changes and @craicoverflow will be adding some changes as part of his pr. Once those are in I will be able to update this correctly.

ui/src/components/AppDetailToolbar.js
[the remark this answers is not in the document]
davidffrench: I was looking into React hooks, which we will not investigate just yet. For now we should be sticking to the high ordered component design pattern where it makes sense.

ui/src/components/AppDetailToolbar.js
davidffrench: perfect

ui/src/components/AppDetailToolbar.js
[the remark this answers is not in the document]
craicoverflow: @laurafitzgerald @davidffrench what is the status of this, can it be reviewed?

craicoverflow:
I can get this to run:

```
./src/components/AppDetailedView.js
Module not found: Can't resolve './AppDetailToolbar' in '/home/endaphelan/go/src/github.com/aerogear/mobile-security-service/ui/src/components'
```

And CircleCI build [fails](https://circleci.com/gh/aerogear/mobile-security-service/1213?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

laurafitzgerald:
@craicoverflow I fixed the issue you saw. It was due to a funky rebase. Should be runnable now.

ui/src/containers/AppDetailsToolbarContainer.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
+import { connect } from 'react-redux';
+import { AppDetailToolbar } from '../components/AppDetailToolbar';
+
+export class AppDetailsToolbarContainer extends React.Component {
+  onDisableButton = () => {
```
craicoverflow: Could be named `onDisableApp` to be more descriptive. It is an event handler which could technically be used by something other that a button.

ui/src/containers/AppDetailsToolbarContainer.js
```diff
@@ -0,0 +1,41 @@
...
+    console.log('Disable button clicked');
+  }
+
+  onHomeClick = () => {
+    this.props.history.push('/');
+  }
+
+  onSaveButton = () => {
```
craicoverflow: Could be named `onSaveApp` to be more descriptive. It is an event handler which could technically be used by something other that a button.

ui/src/components/AppDetailToolbar.js
```diff
@@ -0,0 +1,21 @@
...
+
+export const AppDetailToolbar = ({ onHomeClick, appName, onDisableButton, onSaveButton }) => (
+  <Toolbar className='details-toolbar'>
+    <ToolbarGroup>
+      <Breadcrumb>
+        <BreadcrumbItem onClick={onHomeClick} to='#'>Home</BreadcrumbItem>
+        <BreadcrumbItem isActive >{appName}</BreadcrumbItem>
+      </Breadcrumb>
+    </ToolbarGroup>
+    <ToolbarGroup className='toolbar-buttons'>
```
craicoverflow: There should be a bit of space between the buttons.

ui/src/components/AppDetailToolbar.js
```diff
@@ -0,0 +1,21 @@
...
+      <Button onClick={onDisableButton} variant="primary">Disable App</Button>
```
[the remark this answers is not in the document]
craicoverflow: Actually the buttons on the Patternfly Docs aren't all caps so mabe we should be following their style?

https://patternfly-react.surge.sh/patternfly-4/components/button

ui/src/components/AppDetailedView.js
```diff
@@ -16,6 +16,7 @@ class AppDetailedView extends React.Component {
     return (
       <div className="app-detailed-view">
         <Header />
+        <AppDetailsToolbarContainer />
```
craicoverflow: I think the toolbar would look better if it was the same width as the table below.

ui/src/components/AppDetailToolbar.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Toolbar, ToolbarGroup, Button, Breadcrumb, BreadcrumbItem } from '@patternfly/react-core';
+import { withRouter } from 'react-router-dom';
+import './AppDetailedToolbar.css';
+
+export const AppDetailToolbar = ({ onHomeClick, appName, onDisableButton, onSaveButton }) => (
+  <Toolbar className='details-toolbar'>
+    <ToolbarGroup>
+      <Breadcrumb>
+        <BreadcrumbItem onClick={onHomeClick} to='#'>Home</BreadcrumbItem>
```
craicoverflow: The href here should be "/" not "#" as clicking this will navigate to `http://localhost:3000/#` instead of the expected `http://localhost:3000/`. The navigation history then needs 2 back-clicks to return here.

ui/src/components/common/Header.css
```diff
@@ -1,5 +1,6 @@
-.mssHeader {
+.pf-c-page__header{
```
damienomurchu: @laurafitzgerald I wonder if using the pf internal classes is the right way to go - can you expand on your reasoning for the change?

craicoverflow: Good point @damienomurchu I second that. We should not modify the internal classes, but add a second class to override some of the values.

ui/src/components/AppDetailedToolbar.css
```diff
@@ -0,0 +1,8 @@
+.details-toolbar{
+    margin-left: 3%
```
damienomurchu: Alignment & spacing seems a bit off when I view it, if we can style this to be more consistent with the spacing & alignment of the header, sub-header & table header;

Screenshot from 2019-03-19 11-29-30

ui/src/components/AppDetailedToolbar.css
```diff
@@ -0,0 +1,8 @@
+.details-toolbar{
+    margin-left: 3% 
+}
+.toolbar-buttons{
```
damienomurchu: Styling feedback:

- as already mentioned, some spacing between buttons needs to be added. 
- colour of buttons
  - should both be same colour
  - are chosen colours best to integrate with page

damienomurchu:
@laurafitzgerald good work, basic structure seems to be in place. Some comments left in PR mainly around the styling, as well as +1'ing suggestions made by @craicoverflow

ui/src/components/AppDetailedToolbar.css
laurafitzgerald: @damienomurchu As there is an additional section to go inbetween the breadcrumb and the table I didn't put much thought into the alignment as this may be changed by @craicoverflow's work on that.

ui/src/components/AppDetailedToolbar.css
laurafitzgerald: I've used the default styling which comes with patternfly, i.e. the primary button. We also use the blue buttons in mdc so wanted to keep that consistent. I'm aware that @davidffrench is using a red button in one of the modals so I think we do have some creative licence here.

ui/src/components/AppDetailToolbar.js
laurafitzgerald: The reason I'm doing the route this way is that if I use `/` then the whole page re-renders whereas using `#` only rerenders the content so the navigatioin looks a lot cleaner. In fact the onClick is the logic which is doing the navigation. The reason for including `to='#'` is because it gives the button the correct styling. Without it it doesn't behave like a link.

ui/src/components/AppDetailedView.js
laurafitzgerald: @craicoverflow I will update the padding on the left. In the flat the buttons are more towards the corner so I've left those for now. Let me know if you want that update. Resolving this one for now unless change to the buttons margin is requested.

ui/src/components/AppDetailedToolbar.css
laurafitzgerald: I've made changes to align the breadcrumb with the table so resolving this one for now.

ui/src/components/AppDetailToolbar.js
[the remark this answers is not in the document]
laurafitzgerald: Hmm, I don't think I've done anything to transform it. @davidffrench out of interest what way was the text in the buttons you used for the modal. I can't recall.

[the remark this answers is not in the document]
davidffrench: Patternfly is capitalizing the buttons for me as well. Seems to be the default behaviour, I am not sure why this is not the case in the docs.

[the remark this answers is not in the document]
laurafitzgerald: I found it: I added a text-transform here: https://github.com/aerogear/mobile-security-service/pull/94/files#diff-2e827c03d058d8c099d6ee8dee6693d6R6 This was to uppercase the header of the tables. This can be removed and the text can just be made uppercase in the state.

ui/src/components/AppDetailToolbar.js
craicoverflow: Okay but then it messes up the navigation history so each way has a drawack. Which is better, having the whole view re-rendered and getting the nav history to work as expected, or doing a partial re-render where the user doesn't get what they expect when they press the back button?

We could not include `to='#'` and give it a custom CSS class where we style it ourselves to behave just like a link? I'm sure if we delve into the Patternfly styling we could find it and copy the class.

laurafitzgerald: Yes probably the latter would be the better option as it would satisfy both. I'll have a look into that.

laurafitzgerald:
@damienomurchu @craicoverflow I've updated this pr now after rebasing, addressing feedback and adding tests. It should now be ready for a fresh code review and verification.

ui/src/components/AppDetailsToolbar.js
```diff
@@ -0,0 +1,39 @@
...
+  onSaveApp = () => {
+    console.log('Save button clicked');
+  }
+
+  render () {
+    return (
+      <Toolbar className='details-toolbar'>
+        <ToolbarGroup>
+          <ToolbarItem>
+            <Breadcrumb>
```
damienomurchu: Breadcrumb still looks very small - can it be styled to increase the font a little?

ui/src/components/AppDetailsToolbar.js
```diff
@@ -0,0 +1,39 @@
...
+              <BreadcrumbItem className='home-breadcrumb' onClick={this.onHomeClick}>Home</BreadcrumbItem>
+              <BreadcrumbItem isActive >{this.props.app.appName}</BreadcrumbItem>
+            </Breadcrumb>
+          </ToolbarItem>
+        </ToolbarGroup>
+        <ToolbarGroup className='toolbar-buttons'>
+          <Button className='toolbar-button' onClick={this.onDisableApp} variant="primary">Disable App</Button>
```
damienomurchu: I'm just wondering should both these buttons be the same colour, or should we distinguish them in some way from each other?

ui/src/components/AppDetailsToolbar.js
```diff
@@ -0,0 +1,39 @@
...
+          <Button className='toolbar-button' onClick={this.onSaveApp} variant="primary">Save</Button>
```
damienomurchu: Re spacing on between the buttons, I see patternfly uses the following to add spacing between them, fyi
```
<Button variant="secondary">Secondary</Button>{' '}
<Button variant="tertiary">Tertiary</Button>
```

damienomurchu:
@laurafitzgerald functionally looks ok, still some issues with the styling before it's merged, imo. If this PR is blocking others maybe the styling will be resolved in another PR

ui/src/components/AppDetailsToolbar.js
laurafitzgerald: There was another comment above about this also. Waiting on a reply from @davidffrench about whether we should change button colors. The standard used in mdc is blue so I think let's leave them like that for now so as not to block this pr.

ui/src/components/AppDetailsToolbar.js
laurafitzgerald: Hmm. I've tested that and it doesn't seem to work. Is it necessary to switch from the current implementation?

ui/src/components/AppDetailedToolbar.js
```diff
@@ -0,0 +1,39 @@
...
+import { Toolbar, ToolbarGroup, ToolbarItem, Breadcrumb, BreadcrumbItem, Button } from '@patternfly/react-core';
+import './AppDetailedToolbar.css';
+
+export class AppDetailedToolbar extends React.Component {
+  onDisableApp = () => {
+    console.log('Disable button clicked');
+  }
+
+  onHomeClick = () => {
+    this.props.history.push('/');
```
davidffrench: I would suggest switching this to `this.props.history.goBack();`

laurafitzgerald: using goBack() will navigate the users path rather than back through the tree of the website so that I think we should stick with it as is.

PR #107, opened 2019-03-15
fix: format date (AEROGEAR-8839)

laurafitzgerald:
## Motivation
https://github.com/aerogear/standards/issues/AEROGEAR-8839

## What
Format the date

## Why
Humans can't read. Also it'll contain the timezone

## Verification Steps

1. Render the page and view the date 

## Checklist:

## Progress

- [x] Implement
- [x] Rebase on [this pr](https://github.com/aerogear/mobile-security-service/pull/95)

### DIFF
--- a/ui/src/containers/AppVersionsTableContainer.js
+++ b/ui/src/containers/AppVersionsTableContainer.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { Checkbox, TextInput } from '@patternfly/react-core';
+import moment from 'moment';
 import { getApps, appDetailsSort, updateDisabledAppVersion, updateVersionCustomMessage } from '../actions/actions-ui';
 import AppsTable from '../components/AppsTable';
 import './TableContainer.css';
@@ -59,7 +60,11 @@ export class AppVersionsTableContainer extends React.Component {
       tempRow[0] = versions[i][0];
       tempRow[1] = versions[i][1];
       tempRow[2] = versions[i][2];
-      tempRow[3] = new Date(versions[i][3]).toTimeString();
+      if (versions[i][3].isNullOrUndefined || versions[i][3] === 'Never Launched') {
+        tempRow[3] = 'Never Launched';
+      } else {
+        tempRow[3] = moment(versions[i][3]).format('YYYY-MM-DD HH:mm:ss');
+      }
       tempRow[4] = this.createCheckbox(versions[i][0].toString(), versions[i][4]);
       tempRow[5] = this.createTextInput(versions[i][0], versions[i][5]);
       renderedRows.push(tempRow);
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -3502,8 +3502,7 @@
             },
             "ansi-regex": {
               "version": "2.1.1",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -3521,13 +3520,11 @@
             },
             "balanced-match": {
               "version": "1.0.0",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
-              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
@@ -3540,18 +3537,15 @@
             },
             "code-point-at": {
               "version": "1.1.0",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "concat-map": {
               "version": "0.0.1",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "console-control-strings": {
               "version": "1.1.0",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "core-util-is": {
               "version": "1.0.2",
@@ -3654,8 +3648,7 @@
             },
             "inherits": {
               "version": "2.0.3",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "ini": {
               "version": "1.3.5",
@@ -3665,7 +3658,6 @@
             "is-fullwidth-code-point": {
               "version": "1.0.0",
               "bundled": true,
-              "optional": true,
               "requires": {
                 "number-is-nan": "^1.0.0"
               }
@@ -3678,20 +3670,17 @@
             "minimatch": {
               "version": "3.0.4",
               "bundled": true,
-              "optional": true,
               "requires": {
                 "brace-expansion": "^1.1.7"
               }
             },
             "minimist": {
               "version": "0.0.8",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "minipass": {
               "version": "2.3.5",
               "bundled": true,
-              "optional": true,
               "requires": {
                 "safe-buffer": "^5.1.2",
                 "yallist": "^3.0.0"
@@ -3708,7 +3697,6 @@
             "mkdirp": {
               "version": "0.5.1",
               "bundled": true,
-              "optional": true,
               "requires": {
                 "minimist": "0.0.8"
               }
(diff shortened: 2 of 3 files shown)


laurafitzgerald:
@craicoverflow as I'm working on this one I've noticed that even for items that have a launched value the `Last Launched` column is showing `Never Launched`. I've checked in the DB and the entries do have a value. Are we definitely getting that back from the server?

craicoverflow:
> @craicoverflow as I'm working on this one I've noticed that even for items that have a launched value the `Last Launched` column is showing `Never Launched`. I've checked in the DB and the entries do have a value. Are we definitely getting that back from the server?

Hey @laurafitzgerald you are right. So originally LastLaunchedAt was not getting retrieved from the database at all as it wasn't included in the SELECT query. I put in the "Never Launched" default value because of this, and then I realised it should be returned in the SQL query so I added it. Now the default "Never Launched" is no longer needed! Would you like me to create a new PR or are you doing it here?

laurafitzgerald:
Hi @craicoverflow it was late Friday when I noticed it so I haven't made any changes to address it. I can do it here no problem. It would be easier as this one would get blocked again. Do I need to do anything with the server, restart it again to pull in the change which returns the value?

craicoverflow:
> Hi @craicoverflow it was late Friday when I noticed it so I haven't made any changes to address it. I can do it here no problem. It would be easier as this one would get blocked again. Do I need to do anything with the server, restart it again to pull in the change which returns the value?

Nothing needs to change in the server. You just need to change [this line](https://github.com/aerogear/mobile-security-service/blob/662a8fbc902ecab99552434a08f4f89192fc8748/ui/src/reducers/index.js#L158) to:

```js
temp[3] = version['lastLaunchedAt'];
```

laurafitzgerald:
Tests are failing. Fixing those now.

ui/src/reducers/__tests__/index.test.js
```diff
@@ -135,7 +135,7 @@ describe('reducer', () => {
   ];
 
   it('should return the initial state', () => {
-    expect(reducer(undefined, {})).toEqual(initialState);
+    expect(JSON.stringify(reducer(undefined, {}))).toEqual(JSON.stringify(initialState));
```
damienomurchu: Any reason for the `JSON.stringify` call, or is it just for inspection when testing PR?

ui/src/reducers/index.js
```diff
@@ -27,9 +27,9 @@ const initialState = {
     { title: 'App Version', transforms: [sortable] },
     { title: 'Current Installs', transforms: [sortable] },
     { title: 'Launches', transforms: [sortable] },
-    { title: 'Last Launched', transforms: [sortable] },
+    { title: 'Last Launched', transforms: [sortable, cellWidth(25)] },
```
damienomurchu: Should we add `cellWidth` for all the columns to ensure the proportions remain consistent regardless of screen width?

damienomurchu:
Code looks good, left some comments, but haven't tested. Will verify in the morning

ui/src/reducers/__tests__/index.test.js
laurafitzgerald: The test breaks otherwise. It's a jest issue: https://github.com/facebook/jest/issues/5998 That's the workaround they suggest.

ui/src/reducers/index.js
laurafitzgerald: @damienomurchu I did play with this. The columns default to the size of the column header. As the screen gets smaller on the x axis the columns with cellWidth(25) shrink and wrap text in the cells until the size of the header. Once all columns are as small as to still be able to show the full header the table doesn't shrink any further and the user can then scroll. As far as I can see the cellWidth is maintained only when the screen size will allow. Otherwise the table is responsive to the header text.

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -15,12 +15,14 @@ export class AppVersionsTableContainer extends React.Component {
...
-
       tempRow[0] = versions[i][0];
       tempRow[1] = versions[i][1];
       tempRow[2] = versions[i][2];
-      tempRow[3] = versions[i][3];
-
+      if (versions[i][3].isNullOrUndefined || versions[i][3] === '') {
+        tempRow[3] = 'Never Launched';
+      } else {
+        tempRow[3] = new Date(versions[i][3]).toString();
```
craicoverflow: `Thu Mar 14 2019 16:06:30 GMT+0000 (Greenwich Mean Time)`

Is this the decided upon date format? It is very long.

The date format in the flats look like this:

`2019-04-14 16:06:30`

Which I think is preferable. The end-user should just see the date/time as it is in their timezone and not need to do any calculations to convert from GMT to local time.

Thoughts @davidffrench @damienomurchu ?

davidffrench: I would rather `2019-04-14 16:06:30`. We are saving the date without timezone in the db. So it should always display UTC.

craicoverflow:
@laurafitzgerald after investigation you are no longer blocked. Nothing needs to change on the frontend, you can proceed with this.

https://github.com/aerogear/mobile-security-service/pull/113 I made backend changes here but the way our server is currently set up data will not be changed in any way. This does not need to be merged before your PR can be unblocked.

ui/src/containers/AppVersionsTableContainer.js
damienomurchu: Is there value in displaying the timezone in the date, just so there can be no doubt by the user as to the time? My thinking would be as this is a security service, action may need to be taken in a timely manner, and by displaying the users local timezone in the date the correctness of the date can be ascertained quite quickly

ui/src/reducers/__tests__/index.test.js
damienomurchu: Yes understand. Just hit this issue today & used the same workaround.

ui/src/reducers/index.js
damienomurchu: Understand. Explicit is better than implicit I think, but happy for this to stay the same.

laurafitzgerald:
@craicoverflow thanks for the update. I'll be progressing this one today. I can see that your linked pr is now merged so going to rebase first and progress. Will ping you for a review. :)

laurafitzgerald:
@damienomurchu I see that there is already a change on the date format here: https://github.com/aerogear/mobile-security-service/blob/master/ui/src/containers/AppVersionsTableContainer.js#L62 and on the width of the table columns here: https://github.com/aerogear/mobile-security-service/blob/master/ui/src/reducers/index.js#L31-L36 Are the changes in this branch already covered by those?

damienomurchu:
> @damienomurchu I see that there is already a change on the date format here: https://github.com/aerogear/mobile-security-service/blob/master/ui/src/containers/AppVersionsTableContainer.js#L62 and on the width of the table columns here: https://github.com/aerogear/mobile-security-service/blob/master/ui/src/reducers/index.js#L31-L36 Are the changes in this branch already covered by those?

I believe that change was introduced by @craicoverflow - perhaps he can confirm

craicoverflow:
> > @damienomurchu I see that there is already a change on the date format here: https://github.com/aerogear/mobile-security-service/blob/master/ui/src/containers/AppVersionsTableContainer.js#L62 and on the width of the table columns here: https://github.com/aerogear/mobile-security-service/blob/master/ui/src/reducers/index.js#L31-L36 Are the changes in this branch already covered by those?
> 
> That change was introduced by @craicoverflow - perhaps he can give you the clarification you need

I converted `lastLaunchedAt` prop (which is a string) to a date because otherwise it would remain a string and would not convert to local time from UTC. It probably needs further formatting.

laurafitzgerald:
Waiting for feedback around what format this date should take so leaving the blocked label on this one for now.

laurafitzgerald:
@craicoverflow @damienomurchu this is now ready for review/verification. I'm looking into the option of having a tooltip which shows the amount of time from when the app was last launched but it's not part of this pr so I will create a JIRA if there is a way to do that easily.

laurafitzgerald:
thanks @craicoverflow would you mind creating a task for that. squashed commits and will merge once builds are completed.

PR #108, opened 2019-03-19
fix(table-sorting): Don't sort table if column values are all equal (AEROGEAR-8859)

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8859

## What

Prevent the table from sorting on a column if every value in that column are equal.

## Why

This is expected behaviour. The table should stay the same in this scenario.

## How

Compared all values in column and return the original `rows` array if they are.

## Verification Steps
 
1. Try to sort on a table column where all values are the same (if they aren't, change them to be the same).
2. The table state should remain exactly the same.
3. Sort on a column where values are not the same.
4. The sorting order should change.

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/ui/src/reducers/index.js
+++ b/ui/src/reducers/index.js
@@ -44,7 +44,11 @@ const initialState = {
 // returns a new array sorted in preferred direction
 const sortRows = (rows, index, direction) => {
   // sort in ascending direction
-  const sortedRows = rows.sort((a, b) => (a[index] < b[index] ? -1 : a[index] > b[index] ? 1 : 0));
+  const sortedRows = [...rows].sort((a, b) => (a[index] < b[index] ? -1 : a[index] > b[index] ? 1 : 0));
+
+  if (areColumnValuesEqual(rows, index)) {
+    return rows;
+  }
 
   // reverse if descending direction is preferred
   if (direction !== SortByDirection.asc) {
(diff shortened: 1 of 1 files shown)


ui/src/reducers/index.js
```diff
@@ -74,7 +104,7 @@ export default (state = initialState, action) => {
     case APP_VERSIONS_SORT:
       const versionDirection = action.payload.direction;
       const versionIndex = action.payload.index;
-      const sortedAppVersions = sortRows(state.app.versionsRows, versionIndex, versionDirection);
+      const sortedAppVersions = sortRows(state.app.versionsRows, versionIndex, versionDirection, state.appVersionsSortDirection.index, state.appVersionsSortDirection.direction);
```
laurafitzgerald: @craicoverflow why are these additional values added here?

craicoverflow: woops I put them in when I was testing a way to fix this but forgot to take em out! will remove.

laurafitzgerald:
Functionality verified 👍 but waiting on questions in code review to approve.

ui/src/reducers/index.js
```diff
@@ -54,6 +58,32 @@ const sortRows = (rows, index, direction) => {
   return sortedRows;
 };
 
+/**
+ * Check if all of the table column values are the same
+ *
+ * @param {Array} rows The table rows in which we will compare values
```
laurafitzgerald: @craicoverflow should we be doing this in other methods?

craicoverflow: I was wondering that also and I think we should as it the standard JavaScript way.

Look how clear it makes a function when you hover over it - you know exactly what it does and what the params are for. Documentation can also be generated when this format is used.

Screenshot from 2019-03-19 10-40-28

Ref: [JSDoc](http://usejsdoc.org/)

PR #109, opened 2019-03-19
Aerogear 8467

damienomurchu:
## Motivation
Aerogear 8467

## What
Update client-side model for checkbox & custom message changes

## Why
Aerogear 8467

## How
Update listener logic, add new reducers, add tests

## Verification Steps

Setup:
- ensure db is running `docker-compose up -d db`
- `make serve` 

## Checklist:

## Progress

- [ ] Finished task


ui/src/components/App.js
```diff
@@ -10,7 +10,7 @@ class App extends React.Component {
       <Router>
         <div className="App">
           <Switch>
-            <Route exact path="/" component={LandingPage} />
+            <Route exact path="/" component={AppDetailedView} />
```
damienomurchu: just for purpose of testing pr, will be removed later

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -1,40 +1,70 @@
...
+import { Checkbox, TextInput } from '@patternfly/react-core';
+import { getApps, appDetailsSort, updateDisabledApp } from '../actions/actions-ui';
 import AppsTable from '../components/AppsTable';
 import './TableContainer.css';
 
 export class AppVersionsTableContainer extends React.Component {
+  componentWillMount () {}
+
+  handleCheckedChange = (_event, e) => {
+    console.log('checkbox clicked');
```
damienomurchu: will be removed before merging

ui/src/reducers/index.js
```diff
@@ -17,26 +18,32 @@ const initialState = {
...
+    { title: 'Disable on Startup', transforms: [ sortable ] },
+    { title: 'Custom Disable Message', transforms: [ sortable ] }
   ],
   isAppsRequestFailed: false,
   currentUser: 'currentUser',
   isUserDropdownOpen: false,
   app: {
     data: {},
-    versionsRows: []
+    versionsRows: [
```
damienomurchu: for local testing, will be removed before merging

ui/src/reducers/index.js
```diff
@@ -150,6 +152,40 @@ export default (state = initialState, action) => {
         ...state,
         isUserDropdownOpen: !isUserDropdownOpen
       };
+    case UPDATE_DISABLED_APP:
+      const id = action.payload.id;
+      const disabled = action.payload.disabled;
```
laurafitzgerald: Could we use `isDisabled` rather than `disabled`? This is a boolean value right?

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -1,40 +1,69 @@
...
+
+  createTextInput = (id, text) => {
+    return (
+      <React.Fragment>
+        <TextInput
+          id={id}
+          type="text"
+          placeholder="Add a custom message.."
+          value={text}
+          onChange={this.handleTextInputChange}
```
damienomurchu: @davidffrench I've left the above 2 lines in to be able to check the client model is updating correctly

ui/src/reducers/index.js
damienomurchu: done

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -1,40 +1,69 @@
...
+          aria-label="disable app checkbox"
+          id={id}
+        />
+      </React.Fragment>
+    );
+  };
+
+  createTextInput = (id, text) => {
+    return (
+      <React.Fragment>
```
craicoverflow: You could use a Patternfly 4 [TextArea](http://patternfly-react.surge.sh/patternfly-4/components/textarea).

ui/src/reducers/index.js
```diff
@@ -150,6 +152,38 @@ export default (state = initialState, action) => {
...
+        app: {
+          data: state.app.data,
+          versionsRows: [ ...updatedVersions ]
+        }
+      };
+    case UPDATE_CUSTOM_MESSAGE:
+      const versionId = action.payload.id;
+      const value = action.payload.value;
+      const updatedVersionsRows = state.app.versionsRows.map((version) => {
+        if (version[0] === versionId) {
```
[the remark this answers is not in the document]
damienomurchu: @craicoverflow yes, I think so. index reference could change easily in future, so best to switch to an object & reference via an object key. May not refactor in this PR though, but think it's a worthwhile change and we can do in another PR

ui/src/containers/AppVersionsTableContainer.js
damienomurchu: changed to onBlur now.

ui/src/containers/AppVersionsTableContainer.js
damienomurchu: Yes, I think this would be the solution to allow for multiline comments - let's address this in the PR that wraps this field in a form though I think

ui/src/reducers/index.js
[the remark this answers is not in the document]
damienomurchu: Just looking at this again, and realised patternfly Table component needs an array of row values (which is what we're storing with `versionsRows` in state), so don't think this refactor will be possible unfortunately

damienomurchu:
@davidffrench I think this PR represents what's required in the JIRA. If there's more that needs to be done for the task to be considered complete please shout :)

davidffrench:
Hey @damienomurchu . I have tested the PR and there are a couple of issues.

1. The TextInput is not letting me type characters into it. This seems to be intermittent, I haven't been able to figure out what trigger allows me to start typing characters.

2. Due to what seems to be the onBlur listener, the next action of the user when bluring from the field is cancelled. E.g. Focus a textInput field and click a checkbox, this will not work.

3. The textInput field is too short. The easiest way to solve this is setting a column width on the other columns in the table and leaving the last column as max. This can be seen in the Table with Width Modifiers sample here - http://patternfly-react.surge.sh/patternfly-4/components/table

4. I haven't been able to test the sort of the textInput field to see how it works.

damienomurchu:
@davidffrench updated changes pushed to address points 3 & 4 - let me know if column width for textInput is appropriate (I tried just max, but it seemed too wide to me). As discussed points 1 & 2 to be resolved in related task & PR

PR #111, opened 2019-03-20
feat(ui): Add navigation modal to UI (AEROGEAR-8473)

davidffrench:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8473

## What
Create a navigation modal

## Why
To advise the end user they have unsaved changes before leaving the detailed UI page.

## Verification Steps
1. Checkout this branch locally
2. Modify the isAppDetailedDirty property in the reducer initalState to true
3. Run the server and UI using `make serve`
4. Navigate to the detailed app screen
5. Click the back button in the browser or the header
6. Verify the navigation modal appears
7. Verify the `X` and `Stay` buttons close the modal and keep you on the same screen
8. Verify the `Leave` button navigates back to the landing page

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


ui/src/components/AppDetailedView.js
```diff
@@ -12,6 +14,21 @@ class AppDetailedView extends React.Component {
...
+      if (this.props.isDirty) {
+        this.props.toggleNavigationModal(true);
+        return false;
+      }
+    });
+  }
+
+  componentWillUnmount () {
+    this.unblock();
+    // return true;
```
craicoverflow: Comment should be removed.

ui/src/containers/NavigationModalContainer.js
```diff
@@ -0,0 +1,50 @@
...
+  handleLeaveClick = () => {
+    this.props.toggleAppDetailedIsDirty();
+    this.props.history.goBack();
+    this.handleModalClose();
+  };
+
+  render () {
+    return (
+      <LargeModal
+        title="Are you sure you want to leave this page?"
```
craicoverflow: I am in favour of making components as reusable as possible. To make this component reusable we can pass props that are likely to be different on different pages.

The following could be passes in as `props`:

* title
* text
* Primary button text (Leave)
* Secondary button (Stay)

ui/src/containers/NavigationModalContainer.js
```diff
@@ -0,0 +1,50 @@
...
+        isOpen={this.props.isNavigationModalOpen}
+        onClose={this.handleModalClose}
+        actions={[
+          <Button key="leave" variant="danger" onClick={this.handleLeaveClick}>
```
craicoverflow: Is red the colour we want to go with here? For me, red is reserved for extremely risky actions, such as deleting an app. Yes disabling an app is somewhat risky but is to the same degree?

I'm not sure myself by the way, it is just worth discussing.

davidffrench: I decided to go with red because the user's unsaved data will be lost if they navigate away.

ui/src/containers/NavigationModalContainer.js
davidffrench: You believe the text should be stored in the redux state and passed into this component?

craicoverflow: You don't have to store something in the state in order to pass it as a prop to a component. You can still pass props to your component, like this:

```jsx
return (
    <NavigationModalContainer title="Are you sure you want to leave this page?"/>
);
```

davidffrench: Changed now for the title and text. I don't believe the button text should change as the handlers are very specific to the what they should be doing.

ui/src/components/AppDetailedView.js
```diff
@@ -12,6 +14,20 @@ class AppDetailedView extends React.Component {
     this.props.getAppById(this.props.match.params.id);
   }
 
+  componentDidMount () {
+    this.unblock = this.props.history.block(targetLocation => {
```
craicoverflow: `targetLocation` is not used and can be removed.

PR #119, opened 2019-03-25
feat(ui): Created a modal to make users confirm they wish to save an app (AEROGEAR-8472)

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8472

## What

Create a confirmation modal that will be shown to the user to confirm they wish to save an app.

## Why

It is good practice to make a user confirm an action such as modifying critical server data.

## Verification Steps
 
1. Navigate to the app detailed page.
2. Click `Save App`.
3. A confirmation modal will appear.
## Checklist:

## Progress

- [x] Finished task


ui/src/actions/actions-ui.js
```diff
@@ -8,6 +8,7 @@ import {
   APPS_SORT,
   TOGGLE_HEADER_DROPDOWN,
   TOGGLE_NAVIGATION_MODAL,
+  TOGGLE_CONFIRM_SAVE_APP_MODAL,
```
damienomurchu:
```suggestion
  TOGGLE_SAVE_APP_MODAL,
```

just wonder if we need the `CONFIRM`

ui/src/actions/actions-ui.js
```diff
@@ -52,6 +53,15 @@ export const toggleNavigationModal = (isOpen, targetLocation) => {
   };
 };
 
+export const toggleConfirmationModal = isConfirmSaveAppModalOpen => {
+  return {
+    type: TOGGLE_CONFIRM_SAVE_APP_MODAL,
```
damienomurchu: same as above comment re `CONFIRM`

ui/src/actions/actions-ui.js
craicoverflow: I have it as `CONFIRM` because the modal does not actually execute the logic to save the app, it gives the go ahead to the parent container to do this. So having it as `SAVE_APP_MODAL` could cause one to believe it will be saving the app. I can still change it if it is still unnecessary, but ant to hear your thoughts on my reason for naming it as this.

ui/src/components/AppDetailedToolbar.js
```diff
@@ -12,24 +12,22 @@ export class AppDetailedToolbar extends React.Component {
     this.props.history.push('/');
   }
 
-  onSaveApp = () => {
-    console.log('Save button clicked');
-  }
-
   render () {
+    const { app, onSaveApp, onDisableApp } = this.props;
```
damienomurchu: I see these are coming in from the parent component - can be add PropTypes in this class to add some checks for them? ie

```
AppDetailedView.propTypes = {
  app: PropTypes.object.isRequired,
  onSaveApp: PropTypes.func.isRequired,
  onDisableApp: PropTypes.func.isRequired
};
```

craicoverflow: Yep makes sense.

ui/src/components/AppDetailedView.js
```diff
@@ -29,18 +30,37 @@ class AppDetailedView extends React.Component {
     this.unblockHistory();
   }
 
+  onSaveApp () {
```
damienomurchu:
```suggestion
  onSaveApp = () => {
```
this will bind the `this` to this component when called, otherwise you will have to bind it in a constructor here, ie
```
this.onSaveApp = this.onSaveApp.bind(this);
```

ui/src/components/AppDetailedView.js
```diff
@@ -63,7 +83,8 @@ function mapStateToProps (state) {
 
 const mapDispatchToProps = {
   getAppById,
-  toggleNavigationModal
+  toggleNavigationModal,
```
damienomurchu: I wonder if anything in `mapDispatchToProps` is required, as these functions should all be available as props through the import at the top of the file, ie:
```
import { getAppById, toggleNavigationModal, toggleConfirmationModal } from '../actions/actions-ui';
```

ui/src/containers/ConfirmSaveAppModalContainer.js
```diff
@@ -0,0 +1,47 @@
...
+        confirmAction={[<Button key="confirm" variant="primary" onClick={onConfirm}>
+            Confirm
+        </Button>]}>
+        {children}
+      </ConfirmationModal>
+    );
+  }
+}
+
+ConfirmSaveAppModalContainer.propTypes = {
```
damienomurchu: Should `onConfirm` be here too?

ui/src/components/AppDetailedView.js
craicoverflow: It doesn't seem that I need to that binding in the constructor - what is there currently works.

ui/src/components/AppDetailedView.js
craicoverflow: `toggleConfirmSaveAppModal` does come from Redux.

https://github.com/aerogear/mobile-security-service/blob/b5a132092606f1f17d7673b2b5c669c23ed391d5/ui/src/actions/actions-ui.js#L56

ui/src/actions/actions-ui.js
damienomurchu: Hmm, my own reading of it would be `TOGGLE_SAVE_APP_MODAL` (or `TOGGLE_CONFIRM_SAVE_APP_MODAL`) makes it clear enough it's just a toggling of the modal going on, so proposed rename would just be to reduce verbosity. Either name is good to me though

ui/src/components/AppDetailedView.js
damienomurchu: My mistake, misread `mapDispatchToProps` as `mapStateToProps` - need :coffee:  :laughing:

damienomurchu:
@craicoverflow code looks good to me, one remaining comment left in PR. Will verify functionality shortly

ui/src/containers/ConfirmSaveAppModalContainer.js
craicoverflow: Done.

ui/src/actions/actions-ui.js
craicoverflow: Changed.

ui/src/actions/actions-ui.js
craicoverflow: Changed.

ui/src/components/AppDetailedView.js
damienomurchu: Yes, in this case it seems the context is handled correctly as it's handled by passing the entire handler as a prop to `AppDetailedToolbar`. I would still suggest explicitly binding `this` though to avoid potential future bugs if `onSaveApp` is called from a different context than `AppDetailedView`

craicoverflow: Thanks for the explanation :+1:

craicoverflow: Updated with your suggested change.

ui/src/actions/actions-ui.js
```diff
@@ -52,6 +53,15 @@ export const toggleNavigationModal = (isOpen, targetLocation) => {
   };
 };
 
+export const toggleConfirmSaveAppModal = isConfirmSaveAppModalOpen => {
```
damienomurchu: @craicoverflow hadn't realised the renaming of `ConfirmSaveApp` to `SaveApp` touched so many places. I'd suggest keeping the naming consistent, and either removing or keeping `confirm` in the method and variable names associated with save-app-modal.

PR #120, opened 2019-03-26
fix(ui): Add dynamic targetLocation to NavigationModal (AEROGEAR-8887)

davidffrench:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8887

## What
Changing `goBack` to dynamic `targetLocation` so the Navigation modal will work for any react router history location change.

## Why
Future proof

## How
Passing target location into Redux from the history block function so it can be used by the navigation modal later on.

## Verification Steps
1. Checkout this branch locally
2. Modify the isAppDetailedDirty property in the reducer initalState to true
3. Run the server and UI using `make serve`
4. Navigate to the detailed app screen
5. Click the back button in the browser or the header logo
6. Verify the navigation modal appears
7. Verify the X and Stay buttons close the modal and keep you on the same screen
8. Verify the Leave button navigates back to the landing page

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

PS.: Add images and/or .gifs to illustrate what was changed if this pull request modifies the appearance/output of something presented to the users.

### DIFF
--- a/ui/src/reducers/index.js
+++ b/ui/src/reducers/index.js
@@ -38,7 +38,10 @@ const initialState = {
   isAppsRequestFailed: false,
   currentUser: 'currentUser',
   isUserDropdownOpen: false,
-  isNavigationModalOpen: false,
+  navigationModal: {
+    isOpen: false,
+    targetLocation: undefined
+  },
   isAppDetailedDirty: false,
   app: {
     data: {},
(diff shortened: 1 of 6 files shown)


ui/src/reducers/index.js
```diff
@@ -38,7 +38,10 @@ const initialState = {
   isAppsRequestFailed: false,
   currentUser: 'currentUser',
   isUserDropdownOpen: false,
-  isNavigationModalOpen: false,
+  navigationModal: {
+    isNavigationModalOpen: false,
```
craicoverflow: Since this is a property of `navigationModal`, it could be reduced to just `isOpen`, as it is implied that the prop is referring to the navigation modal.

`navigationModal.isOpen` vs `navigationModal.isNavigationModalOpen`.

damienomurchu:
@davidffrench just verifying the functionality works as expected, and also that `isAppDetailedDirty` gets updated as expected too in the redux store. 

Just skimmed code though, will review properly now and leave any comments.

ui/src/reducers/index.js
davidffrench: Good point and addressed now.

PR #123, opened 2019-03-28
ci(commit-msg): Added allowance for extra types in commit message (AEROGEAR-8888)

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8888

## What

Extended the commit-msg hook to allow extra types to be defined according to the Conventional Commit specification.

Also I noticed that the `commit-msg` hook was checking previous commits but not the current commit. 
Because of this, the current commit message was passing and getting committed even if the format was incorrect. During the next commit the hook was failing on that previous commit. I changed the script to only look at the content of the current commit message.

Allowed types: _feat, fix, docs, breaking, build, ci, perf, refactor, style, test_

## Why

We only had 4 types - docs, feat, fix, breaking. These will probably be used 80% of the time but limiting to these is too restrictive. By adding more we can make our commit messages more descriptive.

## How

Extended the current commit-msg hook.

## Verification Steps

1. Create a temporary branch to make test commits:

```sh
git checkout -b temp-branch
```

2. Make a file change and try to make the following commit:

```sh
git commit -m "invalid format"
```

The commit should fail.

3. Try again with a valid format, with one of the new types:

```sh
git commit -m "refactor(test): refactoring (AEROGEAR-8888)"
```

The commit should pass.

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,8 +1,45 @@
 #!/bin/sh
 
-./scripts/commit-filter-check.sh $1 && rc=$? || rc=$?
-if [ $rc == 1 ]
-then
-	exit 1
-fi
+commit_message_check (){
+      # gets the content of the current commit message
+      gitmessage="$(cat $1)"
+      
+      matchtypes="feat\|fix\|docs\|breaking\|build\|ci\|perf\|refactor\|style\|test"
+      types="feat, fix, docs, breaking, build, ci, perf, refactor, style, test"
+      
+      messagecheck=`echo $gitmessage | grep -w "$matchtypes"`
 
+      if [ -z "$messagecheck" ]
+      then 
+            echo -e "Your commit message must begin with one of the following: [\e[33m$types\033[0m]"
+      fi
+      messagecheck=`echo $gitmessage | grep "(AEROGEAR-"`
+      if  [ -z "$messagecheck" ]
+      then 
+            echo -e "Your commit message must end with the following: \e[33m(AEROGEAR-****)\033[0m where **** is the JIRA number"
+            echo " " 
+      fi
+
+      messagecheck=`echo $gitmessage | grep ": "`
+      if  [ -z "$messagecheck" ]
+      then 
+            echo "Your commit message has a formatting error please take note of special characters '():' position and use in the example below:"
+            echo -e "\n\t\e[34mtype(optional-scope): some txt (AEROGEAR-****)\033[0m"
+            echo " "
+      fi
+
+      messagecheck=`echo $gitmessage | grep -w "$matchtypes" | grep "(AEROGEAR-" | grep ": "`
+
+      # check to see if the messagecheck var is empty
+      if [ -z "$messagecheck" ]
+      then  
+            echo -e "\e[31m[COMMIT MESSAGE FAILED]\033[0m Please review the following:\n"
+            echo $gitmessage
+            echo " "
+            exit 1
+      else
+            echo -e "\e[92m[COMMIT MESSAGE PASSED]\033[0m\n"
+      fi  
+}
+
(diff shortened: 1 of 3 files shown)


camilamacedo86:
I think these types are already defined for eng. It is some standard. I remember that has some thread over it too long ago.  Also, you can do feat(refectory) or feat(ci) and etc. Shows that the new types added are not from the same category. I mean the new types shows more specifications which will be instead of ()

.githooks/commit-msg
```diff
@@ -1,12 +1,45 @@
...
+commit_message_check (){
+      # gets the content of the current commit message
+      gitmessage="$(cat $1)"
+      
+      matchtypes="feat\|fix\|docs\|breaking\|build\|ci\|perf\|refactor\|style\|test"
+      types="feat, fix, docs, breaking, build, ci, perf, refactor, style, test"
+      
+      messagecheck=`echo $gitmessage | grep -w "$matchtypes"`
 
+      if [ -z "$messagecheck" ]
```
davidffrench: @craicoverflow What is the reasoning for removing `commit-filter-check.sh` and adding the logic in here?

austincunningham: agree with David here commit-filter-check.sh is still in use in circle ci 

image

but it logic has not been updated to include the new filters

craicoverflow: @davidffrench @austincunningham I've explained this in the description:

> Also I noticed that the commit-msg hook was checking previous commits but not the current commit.
Because of this, the current commit message was passing and getting committed even if the format was incorrect. During the next commit the hook was failing on that previous commit. I changed the script to only look at the content of the current commit message.

This is why I help the two separate. The existing script will validate all commits, but the Git hook is only used to validate the current commit message.

craicoverflow:
> I think these types are already defined for eng. It is some standard. I remember that has some thread over it too long time ago. Shows that the new types added are not from the same category, I mean the new types show more specifications which will be inside of (), for example; `feat(ci), fix(ci), doc(ci)`
> 
> wdyt @davidffrench and @austincunningham?

@camilamacedo86 

The specification is called [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.3/#specification).

As you can see in the examples below, the types I have added **are** commonly used as part of this specification (it is not a standard).

- [commitizen/conventional-commit-types](https://github.com/commitizen/conventional-commit-types/blob/master/index.json)
- [cheat-sheets/conventional-commits](https://www.cheatography.com/albelop/cheat-sheets/conventional-commits/)

In my opinion something like `feat(ci)` is wrong. `feat` is supposed to be used for actual application features. Something which would not apply in a change to the CI build.

One benefit of using conventional commits is that a change log can be generated from the commits. The change log would only look at commits that start with `feat`, `perf` or `fix`. A change log should not have CI related commits, as well as the other types.

[Our own guidelines](https://github.com/conventional-changelog/conventional-changelog) does not state that we should only use one of `docs`, `feat` or `fix`.

.githooks/commit-msg
davidffrench: This makes sense, I am concerned about the logic existing in two places. Could you explain why the `commit-filter-check.sh` is working correctly for the latest commit through the CI but not from the git hook?

craicoverflow: I know and I considered doing a refactor but didn't in the interest of time. I intended for this to be a quick change :grimacing: I can still look at refactoring it now but not sure if it is a priority.

craicoverflow: `commit-filter-check` looks at committed commits. The Git hook checks a commit **message** before it is committed.

austincunningham: agree this is not a priority

PR #124, opened 2019-03-28
fix(ui): no rerender on text input and save local state

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8698

## What
Store the state of the rows locally once text is entered in the input field or a checkbox is changed
Added version object to make the manipulation of the rows easier.

## Verification Steps
 
1. Run the application in development mode with a populated database
2. Using the Chrome React Plugin, navigate down to AppVersionsTableContainer:
Expand State:

_Expected:_

3. After initial page render: shouldAppUpdate = false
4. After initial page render: updatedVersions exist with an array of row data which matches appVersions: state.app.versionsRows (this can be removed later when we remove rows from state.
5. Make a change to any of the Input boxes, this should be reflected in the updatedVersions object
6. Make a change to any of the checkboxes, this should be reflected in the updatedVersions object

7. The screen should not rerender on each local state change and as such the text in the text box is visible and the text box doesn't lose focus.

## Checklist:

## Additional Notes

@craicoverflow I thought that there wasn't enough progress in this to make it a seperate pr but maybe take a look and see if we could use this as is or with some changes. Let's chat about it in the morning.

### DIFF
--- a/ui/src/containers/AppVersionsTableContainer.js
+++ b/ui/src/containers/AppVersionsTableContainer.js
@@ -42,15 +42,15 @@ export class AppVersionsTableContainer extends React.Component {
   handleCustomMessageInputChange = (_event, e) => {
     const id = e.target.id;
     const value = e.target.value;
-    this.state.updatedVersions.forEach(version => {
+    this.state.updatedVersions.map(version => {
       if (version.getVersion() === id) {
         version.setDisabledMessage(value);
       }
+      return version;
     });
     this.setState({
       ...this.state,
-      shouldAppUpdate: false,
-      updatedVersions: this.state.updatedVersions
+      shouldAppUpdate: false
     });
   };
 
@@ -79,7 +79,7 @@ export class AppVersionsTableContainer extends React.Component {
           id={id}
           type="text"
           placeholder="Add a custom message.."
-          value={text === null ? undefined : text}
+          defaultValue={text === null ? undefined : text}
           onChange={this.handleCustomMessageInputChange}
           aria-label="Custom Disable Message"
         />
@@ -89,7 +89,7 @@ export class AppVersionsTableContainer extends React.Component {
 
   getTable = (versions = []) => {
     const renderedRows = [];
-    versions.forEach(version => {
+    versions.map(version => {
       const tempRow = [];
       tempRow[0] = version.getVersion();
       tempRow[1] = version.getAppLaunches();
@@ -103,6 +103,7 @@ export class AppVersionsTableContainer extends React.Component {
       tempRow[4] = this.createCheckbox(version.getVersion(), version.isDisabled());
       tempRow[5] = this.createTextInput(version.getVersion(), version.getDisabledMessage());
       renderedRows.push(tempRow);
+      return version;
     });
     return (
 
--- a/ui/src/containers/AppVersionsTableContainer.js
+++ b/ui/src/containers/AppVersionsTableContainer.js
@@ -3,22 +3,55 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { Checkbox, TextInput } from '@patternfly/react-core';
 import moment from 'moment';
-import { getApps, appDetailsSort, updateDisabledAppVersion, updateVersionCustomMessage } from '../actions/actions-ui';
+import { getApps, appDetailsSort } from '../actions/actions-ui';
 import AppsTable from '../components/AppsTable';
 import './TableContainer.css';
 import config from '../config/config';
 
 export class AppVersionsTableContainer extends React.Component {
+  shouldComponentUpdate = () => {
+    return this.state.shouldAppUpdate;
+  }
+  componentWillMount = () => {
+    this.setState({
+      ...this.state,
+      shouldAppUpdate: false
+    });
+  }
+  componentDidMount = () => {
+    this.setState({
+      ...this.state,
+      updatedVersions: this.props.appVersions
+    });
+  }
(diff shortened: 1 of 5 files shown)


ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -107,15 +139,14 @@ function mapStateToProps (state) {
   return {
     sortBy: state.appVersionsSortDirection,
     columns: state.appVersionsColumns,
-    appVersions: state.app.versionsRows
+    appVersions: state.app.versionsRows,
```
laurafitzgerald: we can probably remove this now and do the translation to the rows and the sorting here in the component.

ui/src/models/versions.js
```diff
@@ -0,0 +1,48 @@
+export default class Version {
```
davidffrench: If we add in a version model, we should also stay consistent and add a model for apps as well.

craicoverflow: Is there a particular usecase/reason to use a model?

laurafitzgerald: I find it a bit cleaner than referencing array[x], particularly as we are now going to be setting the values for the disabled boolean and the disabled message. That's really the only value we get from it so we don't _have_ to use it.

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -56,21 +89,23 @@ export class AppVersionsTableContainer extends React.Component {
 
   getTable = (versions = []) => {
     const renderedRows = [];
-    for (let i = 0; i < versions.length; i++) {
+    versions.map(version => {
```
damienomurchu:
```suggestion
    versions.forEach(version => {
```
(`forEach` makes more sense seeing as we're not making use of the returned array that `.map` creates)

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -56,21 +89,23 @@ export class AppVersionsTableContainer extends React.Component {
...
+        tempRow[3] = moment(lastLaunched).format(config.dateTimeFormat);
       }
-      tempRow[4] = this.createCheckbox(versions[i][0].toString(), versions[i][4]);
-      tempRow[5] = this.createTextInput(versions[i][0], versions[i][5]);
+      tempRow[4] = this.createCheckbox(versionNum, version.isDisabled);
+      tempRow[5] = this.createTextInput(versionNum, version.disabledMessage);
       renderedRows.push(tempRow);
-    }
-
+      return version;
```
damienomurchu: as per above, this unused return can be removed by changing implementation from `.map` to `forEach`

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -3,22 +3,55 @@ import PropTypes from 'prop-types';
...
 import AppsTable from '../components/AppsTable';
 import './TableContainer.css';
 import config from '../config/config';
 
 export class AppVersionsTableContainer extends React.Component {
+  shouldComponentUpdate = () => {
+    return this.state.shouldAppUpdate;
+  }
+  componentWillMount = () => {
+    this.setState({
```
damienomurchu: 2 points;
- if tracking local state in the component I think we should use React hooks to be consistent with other scheduled refactoring that will see us use React hooks to manage local state in our components
- if using `setState`, I think we should be using a constructor to declare the initial state (react docs seem unclear if this is necessary to manipulate state in the class, but it would appear to be good practice in any event)

ui/src/models/versions.js
```diff
@@ -0,0 +1,12 @@
...
+  constructor (json = {}) {
+    this.id = json.id;
+    this.versionNum = json.version;
+    this.appId = json.appId;
+    this.isDisabled = json.disabled;
+    this.disabledMessage = json.disabledMessage;
+    this.numOfAppLaunches = json.numOfAppLaunches;
+    this.currentInstalls = json.numOfCurrentInstalls;
+    this.lastLaunchedAt = json.lastLaunchedAt;
```
damienomurchu: If using a model should we use the default values we currently do for some of these fields?

laurafitzgerald:
Closing as @craicoverflow will take any useful changes to another pf.

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -3,22 +3,55 @@ import PropTypes from 'prop-types';
...
+      ...this.state,
+      updatedVersions: this.props.appVersions
+    });
+  }
   handleDisableAppVersionChange = (_event, e) => {
     const id = e.target.id;
     const isDisabled = e.target.checked;
-    this.props.updateDisabledAppVersion(id, isDisabled);
+    this.state.updatedVersions.forEach(version => {
```
damienomurchu: This implementation will loop through every element in the array each time which seems wasteful give we're just changing on element - can we move towards an implementation more like the functional ones we have in our reducers when we return new state?

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -3,22 +3,55 @@ import PropTypes from 'prop-types';
...
+      shouldAppUpdate: false,
+      updatedVersions: this.state.updatedVersions
+    });
   };
 
   handleCustomMessageInputChange = (_event, e) => {
     const id = e.target.id;
     const value = e.target.value;
-    this.props.updateVersionCustomMessage(id, value);
+    this.state.updatedVersions.map(version => {
```
damienomurchu: same as above re implementation

PR #127, opened 2019-04-01
feat(ui): Add disable app modal (AEROGEAR-8469)

davidffrench:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8469

## What
Adding view for app disable modal

## Why
Allows the user to disable all current versions for an app

## Verification Steps
1. Navigate to the app detailed page.
2. Click `Disable App`.
3. A modal will appear.
4. Enter text in the text field

## Checklist:

## Progress

- [ ] Finished task
- [ ] TODO


ui/src/components/AppDetailedToolbar.js
```diff
@@ -24,7 +24,7 @@ export class AppDetailedToolbar extends React.Component {
         </ToolbarGroup>
         <ToolbarGroup className='toolbar-buttons'>
           <Button className='toolbar-button' onClick={onDisableApp} variant="primary">Disable App</Button>
-          <Button className='toolbar-button' onClick={onSaveApp} variant="primary">Save</Button>
+          <Button className='toolbar-button' onClick={() => onSaveApp(true)} variant="primary">Save</Button>
```
damienomurchu: Should we move the listener implementation out into it's own method? 

Implication of defining it here is that it will be defined every time the component re-renders. Placing it in it's own method in the component means the handler will be re-used instead of being recreated each time.

ui/src/containers/DisableAppModalContainer.js
```diff
@@ -0,0 +1,67 @@
...
+    // this.props.toggleDisableAppModal();
+  };
+
+  const [disableMessage, setDisableMessage] = useState('');
+
+  return (
+    <Modal
+      isLarge
+      title="Disable All App Versions"
+      isOpen={isOpen}
```
damienomurchu:
```suggestion
      isOpen={this.props.isOpen}
```
(not sure you can access it without `this.props`)

ui/src/containers/DisableAppModalContainer.js
```diff
@@ -0,0 +1,67 @@
...
+        <StackItem isMain>
+          <Form isHorizontal>
+            <FormGroup
+              label="Disabled Message"
+              fieldId="horizontal-form-disable-message"
+              helperText="Optional Field. This will overwrite existing custom disabled messages for all versions"
+            >
+              <TextInput
+                value={disableMessage}
+                onChange={(value) => setDisableMessage(value)}
```
damienomurchu: Suggest handler be extracted to its own method for performance reasons when component re-renders

ui/src/containers/DisableAppModalContainer.js
```diff
@@ -0,0 +1,67 @@
...
+        </StackItem>
+        <StackItem isMain>
+          <Form isHorizontal>
+            <FormGroup
+              label="Disabled Message"
+              fieldId="horizontal-form-disable-message"
+              helperText="Optional Field. This will overwrite existing custom disabled messages for all versions"
+            >
+              <TextInput
+                value={disableMessage}
```
damienomurchu: Can't see where `{disableMessage}` is being defined - should this be coming in through a prop somewhere?

ui/src/containers/NavigationModalContainer.js
```diff
@@ -20,7 +19,8 @@ class NavigationModalContainer extends React.Component {
 
   render () {
     return (
-      <LargeModal
+      <Modal
+        isLarge
         title={this.props.title}
         isOpen={this.props.isOpen}
         onClose={this.handleModalClose}
```
damienomurchu: Add `onClose` to required props?

ui/src/reducers/__tests__/index.test.js
```diff
@@ -48,7 +35,12 @@ describe('reducer', () => {
       data: {},
       versionsRows: []
     },
-    isAppRequestFailed: false
+    isAppRequestFailed: false,
+    modals: {
+      disableApp: {
+        isOpen: false
```
damienomurchu: Do we need this to be persisted to the global state, or can it be handled via local state?

ui/src/components/AppDetailedToolbar.js
davidffrench: This is interesting. Do you have any reading material so I can find out more of the pros/cons of what you are suggesting?

ui/src/containers/DisableAppModalContainer.js
davidffrench: On line 7 it is defined. From what I have found, this is the standard way of listing props in a functional component

ui/src/containers/DisableAppModalContainer.js
davidffrench: No prop, this is currently local state using the state hook on line 13

ui/src/containers/DisableAppModalContainer.js
davidffrench: Interested to see what the performance implications of this would be, if you have any materials on it.

ui/src/containers/NavigationModalContainer.js
davidffrench: I don't believe the onClose should be passed in to any of the modals. The modals themselves should be taking care of calling the redux action. I will include this in a larger refactor PR this week.

ui/src/reducers/__tests__/index.test.js
davidffrench: I would prefer to persist in local but this is not possible as the state needs to change from a sibling view (AppDetailedToolbar) and this is exactly what we should be using Redux for.

ui/src/components/AppDetailedToolbar.js
damienomurchu: Actually I was slightly incorrect above - what happens is a new callback is created each time, which if you're passing that callback as a prop to children components can cause them to re-render. That's probably not the case here, as we're just passing it to a PF4 Button component (which probably isn't composed of other children components). 

Snippet from [React docs](https://reactjs.org/docs/handling-events.html):

```
The problem with this syntax is that a different callback is created each time the LoggingButton renders. In most cases, this is fine. However, if this callback is passed as a prop to lower components, those components might do an extra re-rendering. We generally recommend binding in the constructor or using the class fields syntax, to avoid this sort of performance problem.
```

davidffrench: Fantastic, thank you @damienomurchu . This is good to know. I will be conscious of it during the refactoring to ensure it is not passed as a prop to child components.

ui/src/actions/actions-ui.js
```diff
@@ -1,25 +1,10 @@
...
-  APPS_SORT,
-  TOGGLE_HEADER_DROPDOWN,
-  TOGGLE_NAVIGATION_MODAL,
-  TOGGLE_SAVE_APP_MODAL,
-  TOGGLE_APP_DETAILED_IS_DIRTY,
-  APP_VERSIONS_SORT,
-  UPDATE_DISABLED_APP,
-  UPDATE_VERSION_CUSTOM_MESSAGE
-} from '../actions/types.js';
+import * as actionTypes from '../actions/types.js';
```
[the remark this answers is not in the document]
davidffrench: This is true.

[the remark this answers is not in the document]
davidffrench: changed now

davidffrench:
Thanks for the feedback @damienomurchu and @craicoverflow . Could you please review this again. The only additional changes were to fix the issue raised by Enda and change the name of the actions variable.

ui/src/reducers/__tests__/index.test.js
damienomurchu: gotcha, yes, this is probably neater than the alternative (passing event handlers from the parent to the child components)

ui/src/components/AppDetailedToolbar.js
damienomurchu: Yeah, its one of those small potential gotcha's that probably justifies not using inline listeners if only so that you do not need to monitor for edge cases

davidffrench: perfect, I will modify this on the refactor. are you happy to approve the PR?

damienomurchu: Sure, let me just quickly verify the functionality & will approve then :+1:

PR #128, opened 2019-04-01
feat(REST API): Add new endpoint to binding the app (AEROGEAR-8913)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8913

## What
- Create the endpoint to binding the app
- The binding endpoint should be able to create a new app when it does not exist, setup deleted_at as NULL when it has value and updates the name. 

## Why
The operator will call this endpoint to do the binding/re-binding and update the data of some app. 

## How
- New endpoint
- Some changes in the Binding service
- Add new Repository func to update the name.

## Verification Steps

- [ ] check if the CI validations passed
- [ ] check if the endpoint will create a new app
- [ ] check if the endpoint will update the name of some existent app
- [ ] check if the endpoint will set deleted_at NULL when it has a value for the app_id informed. 

Following the endpoint.
* Method: POST
* Path: `/apps/bind/:appId` (The name is optional)
* Body: `{"appName": "Test App"}`

## Checklist:

## Progress

- [X] Finished task
- [ ] TODO

## Additional Notes
Following the tests performed locally.
- [x] check if the endpoint will create a new app
* Informing the name
1. Call the endpoint with the appID POST `localhost:3000/api/apps/bind/com.aerogear.test2` and the body `{"appName": "Test App camila"}`

2. Check if the app was created by GET `localhost:3000/api/apps`

```json
    {
        "id": "714adca7-77ac-48c4-a289-674538c99f85",
        "appId": "com.aerogear.test2",
        "appName": "Test App camila",
        "numOfDeployedVersions": 0,
        "numOfCurrentInstalls": 0,
        "numOfAppLaunches": 0
    },
```

* Without a body ( the name should not be mandatory )
1. Call the endpoint with the appID POST `localhost:3000/api/apps/bind/com.aerogear.test5` and body as `{}`
2. Check if the app was created by GET `localhost:3000/api/apps`

```json
    {
        "id": "d579a68b-5111-4db3-9b6d-fe6156f3c5e7",
        "appId": "com.aerogear.test5",
        "numOfDeployedVersions": 0,
        "numOfCurrentInstalls": 0,
        "numOfAppLaunches": 0
    },
```

- [x] check if the endpoint will update the name of some existent app

1. Call the endpoint with the appID POST `localhost:3000/api/apps/bind/com.aerogear.test2` and the body `{"appName": "New Name"}`

2. Check if the app name was updated by GET `localhost:3000/api/apps`

```json
{
        "id": "714adca7-77ac-48c4-a289-674538c99f85",
        "appId": "com.aerogear.test2",
        "appName": "New Name",
        "numOfDeployedVersions": 0,
        "numOfCurrentInstalls": 0,
        "numOfAppLaunches": 0
    },
```

- [x] check if the endpoint will set deleted_at NULL when it has a value for the app_id informed. 
1. Call the endpoint and/or disabled an app by updating the deleted_at
2. Call the endpoint with the appID POST `localhost:3000/api/apps/bind/:appID` and the body `{}`
3. Check if the app is returned by GET `localhost:3000/api/apps`

### DIFF
--- a/pkg/web/router/router.go
+++ b/pkg/web/router/router.go
@@ -128,6 +128,34 @@ func SetAppRoutes(r *echo.Group, appsHandler apps.HTTPHandler) {
 	//   404:
 	//     description: App not found
 	r.POST("/apps/:id/versions/disable", appsHandler.DisableAllAppVersionsByAppID)
+
+	// Binding/Re-Binding an app
+	// ---
+	// summary:
+	// - Create an new app
+	// - Update deleted_at as NULL to re-binding the app if the app informed be disabled
+	// - Update the name if the name informed is not "" and is different of the name stored in the database for the appId
+	// operationId: BindingApp
+	// produces:
+	// - application/json
+	// parameters:
+	// - name: appId
+	//   in: path
+	//   description: The appId of the app
+	//   required: true
+	//   type: string
+	// - name: body
+	//   in: body
+	//   description:
+	//   required: true
+	//   schema:
+	//     $ref: '#/definitions/App'
+	// responses:
+	//   200:
+	//     description: successful operation
+	//   400:
+	//     description: Invalid id supplied
+	r.POST("/apps/bind/:appId", appsHandler.BindingApp)
 }
 
 func SetInitRoutes(r *echo.Group, initHandler *initclient.HTTPHandler) {
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -11,6 +11,9 @@ definitions:
       appName:
         type: string
         x-go-name: AppName
+      deletedAt:
+        type: string
+        x-go-name: DeletedAt
       deployedVersions:
         items:
           $ref: '#/definitions/Version'
--- a/pkg/models/app.go
+++ b/pkg/models/app.go
@@ -5,9 +5,10 @@ package models
 type App struct {
 	ID                    string     `json:"id"`
 	AppID                 string     `json:"appId"`
-	AppName               string     `json:"appName"`
+	AppName               string     `json:"appName,omitempty"`
 	NumOfDeployedVersions *int       `json:"numOfDeployedVersions,omitempty"`
 	NumOfCurrentInstalls  *int       `json:"numOfCurrentInstalls,omitempty"`
 	NumOfAppLaunches      *int       `json:"numOfAppLaunches,omitempty"`
 	DeployedVersions      *[]Version `json:"deployedVersions,omitempty"`
+	DeletedAt             string     `json:"deletedAt,omitempty"`
 }
--- a/pkg/web/apps/apps_http_handler.go
+++ b/pkg/web/apps/apps_http_handler.go
@@ -17,6 +17,7 @@ type (
 		UpdateAppVersions(c echo.Context) error
 		DisableAllAppVersionsByAppID(c echo.Context) error
 		HealthCheck(c echo.Context) error
+		BindingApp(c echo.Context) error
 	}
 
 	// httpHandler instance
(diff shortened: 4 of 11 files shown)


pkg/web/router/router.go
```diff
@@ -128,6 +128,34 @@ func SetAppRoutes(r *echo.Group, appsHandler apps.HTTPHandler) {
...
+	//   description:
+	//   required: true
+	//   schema:
+	//     $ref: '#/definitions/App'
+	// responses:
+	//   200:
+	//     description: successful operation
+	//   400:
+	//     description: Invalid id supplied
+	r.POST("/apps/bind/:appId", appsHandler.BindingApp)
```
wei-lee: -1 on calling it binding/unbinding. The "bind" term will be used by the MDC, but it should not be used here in the service itself. The end point in the service should just be called for what it does. For example, in this case, might be just call it create app, or reuse the `POST` method for the `/apps` endpoint.

camilamacedo86: @wei-lee the problem is:
- The rest needs to create a new app
- Do the rebinding ( update deleted_at with NULL)
- Update the name of the app in the case of it be changed. 

So, we cannot call it like "createApp" because it also update the app. So, what name would you suggest?

wei-lee: Follow the RESTFul convention, can we not use `POST` to create the app, and `PATCH` to update it, as part of the `apps` endpoint?

camilamacedo86: @wei-lee could you clarify your question and point? 

IMHO in this case, the POST is the most appropriated since it is not following the CRUD standard and REST specs as described above and it will not just update specific data to use the PATCH. It will also create the entity. However, in POV the correct implementation should be we have the GET, PUT, POST endpoint for the entity which is not possible without side effects at this point.  

In this way, please let me know your decision. 

1. Should we move forward with option B which shows less effort now? 
2. Still making sense for you change the name? If yes, what name would you like to represent update the field name and deleted_at, also create new apps? 
3. Should we review the endpoints and make them follow the REST spec in order to be reused for the UI? I mean we change the implementations for we have the GET, PUT, POST endpoint for the entity.
4. Have you any third option?

davidffrench: The endpoints should be named like this which is already sort of discussed above. All we are doing is manipulating the already defined apps resource in our server.

### On appSecurityBinding CR created
POST - /api/apps
This endpoint following the rest convention will allow you to create a new resource for that resource collection.

### On appSecurityBinding CR deleted
PUT or technically PATCH - /api/apps/:id
**Issue:** We do not know the id of the app in the operator which is an issue. 
**Solution:** Following REST, it is perfectly fine to do a PATCH on `/api/apps`. This should still be with the id but we will just have to use the information that we have.

wei-lee: @camilamacedo86 why do we need a specific endpoint for MDC? I think you think a lot about the business logic, but didn't take into account what's it like for end users. From the consumer's POV, the endpoint will create a new app, yes, there are some side effects when the app has been deleted before, but quite honestly, I don't know how many users will actually unbind & bind again. So why not just reuse `POST` to `/api/apps/` or `/api/apps/:appId` to create the app to keep it simple? Maybe I miss something here but what's the benefit of having another endpoint while the current one can be used for the same purpose? 

If we really want to use a different endpoint, I would much prefer it is something like `/api/apps/:appId/enable` because for anyone who hasn't worked with MDC before, it will very clear what the endpoint is for: enable the app. But people will get confused by using `bind`.

camilamacedo86: Hi @wei-lee,

I understand that for us just the GET and PATCH should be enough for us since we do not update the full collection. However, as described above we cannot do that because of the current implementation using the endpoint `api/apps/:id` which is not following the REST standards. 

**Suggestion/Solution**

- Move the current impl GET `api/apps/:id` for `api/apps/:id/details` since it is returning specifically a content which is the details of the app valid just for the UI
- Create the GET  `api/apps/:appId` - Return the App (struct)
- Create the PATCH `api/apps/:appId` - To update the deleted_at and name which are the only fields of this app that we can update. 

Wdyt?

@austincunningham @davidffrench @damienomurchu @laurafitzgerald -> Could we discuss it here or you would like that it be raised for email? Have you any objection for we do this refectory? 

IHMO is better to do it now instead of to propagate this anti-partner and start to create/impl many specific endpoints for each future need of this project. Note that this impl is leading us to start to impl endpoints like `/api/apps/createApp` and `updateNameApp` as this one bind and unbinding when in the fact if we follow the REST standard we will have the GET, DELETE, PUT, POST, PATCH app and the MDC/UI as any future service that needs use this REST API will be allowed to do it as should be.  

**References:** 

- https://medium.com/hashmapinc/rest-good-practices-for-api-design-881439796dc9
- https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design
- https://hackernoon.com/restful-api-designing-guidelines-the-best-practices-60e1d954e7c9
- https://restfulapi.net/http-methods/

camilamacedo86: After spoke with @wei-lee was defined:

For now we can create the POST `/api/apps/:appId` to move foward and solve the design problem with the GET `/api/apps/:id` later.

camilamacedo86: After spoke with the @davidffrench following what was defined to be done to move forward now. 

Definitions:

- GET /apis/apps/?appId=
- POST /apis/apps ( request body =  {App def. appId and appName} / response body empty 201 ) (Update/Create)
- DELETE /apis/apps/:id  -> will set the date and time in deleted_at

Also, I will do this changes in a new PR. Really tks for all your support folks.

PR #129, opened 2019-04-02
fix: Prevent input losing focus after every character entered (AEROGEAR-8918)

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8918

## What

Changed the input event handler to use `onBlur` to update the state in the Redux store.

## Why

When the input used `onChange`, the state updated after every entered character, initiating a full re-render and making the input lose focus.

## Verification Steps
 
1. Go to the app detailed view - `Home > Test App`.
2. For one of the versions in the table, type characters. You should be able to continuously type a stream of characters.
3. Click away from the input. 
4. Inspect the updated state in your browser console. 
5. Look at `app.versionsRows`. Do you see the updated value that matches whatever you have typed in your textbox?

## Checklist:

## Progress

- [ ] Finished task
 
## Additional Notes

You will see the following warnings in the browser console:

```
Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components
```

```
index.js:1446 Warning: TextInput contains an input of type text with both value and defaultValue prop
```

I think this is unavoidable in our current implementation. 

First off, we are using Patternfly 4's [`TextInput`](http://patternfly-react.surge.sh/patternfly-4/components/textinput). This component wraps the core React input, and automatically sets the `value` property to null if it is not set by us. If we were using the core React input component then we would not be setting `value` at all and this warning would not appear.

The definition of `defaultValue` is as follows:

> The defaultValue property sets or returns the default value of a text field.

Ideally we should not be using this property at all, but it is needed as part of a workaround to prevent this  **dynamically created input*** from losing focus each time a character is entered. There is a refactor of the Redux implementation in progress, and I believe this should be revisited when that is complete to see if we can use `value`.

*_This issue does not seem to happen on lone text inputs that are not dynamically created._

### DIFF
--- a/ui/src/containers/AppVersionsTableContainer.js
+++ b/ui/src/containers/AppVersionsTableContainer.js
@@ -9,20 +9,48 @@ import './TableContainer.css';
 import config from '../config/config';
 
 export class AppVersionsTableContainer extends React.Component {
+  state = {
+    canUpdateComponent: true
+  }
+
+  shouldComponentUpdate () {
+    return this.state.canUpdateComponent;
+  }
+
+  componentDidUpdate () {
+    this.setState({
+      canUpdateComponent: false
+    });
+  }
+
   handleDisableAppVersionChange = (_event, e) => {
     const id = e.target.id;
     const isDisabled = e.target.checked;
-    this.props.updateDisabledAppVersion(id, isDisabled);
+
+    this.setState({
+      canUpdateComponent: false
+    }, () => {
+      this.props.updateDisabledAppVersion(id, isDisabled);
+    });
   };
 
-  handleCustomMessageInputChange = (_event, e) => {
+  handleCustomMessageInputChange = (e) => {
     const id = e.target.id;
     const value = e.target.value;
-    this.props.updateVersionCustomMessage(id, value);
+
+    this.setState({
+      canUpdateComponent: false
+    }, () => {
+      this.props.updateVersionCustomMessage(id, value);
+    });
   };
 
   onSort = (_event, index, direction) => {
-    this.props.appDetailsSort(index, direction);
+    this.setState({
+      canUpdateComponent: true
+    }, () => {
+      this.props.appDetailsSort(index, direction);
+    });
   };
 
   createCheckbox = (id, checked) => {
@@ -46,8 +74,8 @@ export class AppVersionsTableContainer extends React.Component {
           id={id}
           type="text"
           placeholder="Add a custom message.."
-          value={text}
-          onChange={this.handleCustomMessageInputChange}
+          defaultValue={text}
+          onBlur={this.handleCustomMessageInputChange}
           aria-label="Custom Disable Message"
         />
       </React.Fragment>
@@ -94,7 +122,7 @@ export class AppVersionsTableContainer extends React.Component {
       );
     }
 
-    return this.getTable(this.props.appVersions);
+    return this.getTable([...this.props.appVersions]);
   }
 }
 


davidffrench:
@craicoverflow Unfortunately there are a couple of issues I found from verifying the issue.

### Console errors
I am seeing the following errors in the console
```
Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```

``` 
TextInput contains an input of type text with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled input element and remove one of these props. More info: https://fb.me/react-controlled-components
```

### Only versionRows updated in Redux
The versionRows are getting updated in Redux. I would have also expected the state to be updated. The versions needed to send to the server I would assumed should come from data.

### The view still rerenders on blur
This means the action performed on blur is ignored. to reproduce this. Focus the field and then click on a checkbox. This click action is ignored.

craicoverflow:
> @craicoverflow Unfortunately there are a couple of issues I found from verifying the issue.
> 
> ### Console errors
> I am seeing the following errors in the console
> 
> ```
> Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
> ```
> ```
> TextInput contains an input of type text with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled input element and remove one of these props. More info: https://fb.me/react-controlled-components
> ```
> ### Only versionRows updated in Redux
> The versionRows are getting updated in Redux. I would have also expected the state to be updated. The versions needed to send to the server I would assumed should come from data.
> 
> ### The view still rerenders on blur
> This means the action performed on blur is ignored. to reproduce this. Focus the field and then click on a checkbox. This click action is ignored.

### Console errors

I will talk to you about this with @laurafitzgerald 

#### Only versionRows updated in Redux

This is a fix of an existing implementation. It is outside of the scope of this PR to change what is updated. Again I will speak to you about this.

### The view still rerenders on blur

Hmm...good spot.

damienomurchu:
Just to confirm I'm getting the same errors as noted by @davidffrench above

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -9,20 +9,42 @@ import './TableContainer.css';
 import config from '../config/config';
 
 export class AppVersionsTableContainer extends React.Component {
+  state = {
+    canUpdateComponent: true
+  }
+
+  shouldComponentUpdate () {
+    return this.state.canUpdateComponent;
```
damienomurchu: are we using this correctly? docs & examples I'm looking at tend to look like:
```
shouldComponentUpdate(nextProps, nextState) {
  return nextProps.isFavourite != this.props.isFavourite;
}
```
Links I was reading: https://reactjs.org/docs/react-component.html#shouldcomponentupdate, https://medium.freecodecamp.org/react-shouldcomponentupdate-demystified-c5d323099ef6

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -9,20 +9,42 @@ import './TableContainer.css';
 import config from '../config/config';
 
 export class AppVersionsTableContainer extends React.Component {
+  state = {
```
damienomurchu: if using local state like this, do we need a constructor?

ui/src/containers/AppVersionsTableContainer.js
```diff
@@ -9,20 +9,42 @@ import './TableContainer.css';
...
+
   handleDisableAppVersionChange = (_event, e) => {
     const id = e.target.id;
     const isDisabled = e.target.checked;
-    this.props.updateDisabledAppVersion(id, isDisabled);
+
+    this.setState({
+      canUpdateComponent: false
+    }, () => {
+      this.props.updateDisabledAppVersion(id, isDisabled);
```
damienomurchu: I think this is definitely the correct approach to handling this, just calling the reducer action from the onBlur event

damienomurchu:
@craicoverflow approach looks good, need to look into the errors we're seeing in the console, as seems a bad smell if we're seeing warnings/ errors in our dev environment.

What I notice when I update the version message field and confirm the save is that the change does not persist. I presume this is because although the redux local state is getting updated when I add the message, that state is then getting overwritten when the API call goes through when the screen re-renders. Have you been able to confirm this is why the updated message is not persisting in the view?

ui/src/containers/AppVersionsTableContainer.js
craicoverflow: There is no constructor here.

damienomurchu: Yes, but shouldn't we be initialising local state in our class component? If so, we [should need](https://reactjs.org/docs/react-component.html#constructor) a constructor to do so

ui/src/containers/AppVersionsTableContainer.js
craicoverflow: I first tried this way and I don't think it would work for what we are trying to do. First off, we are dealing with a dynamic list of versions, where two properties from each can be updated. We could either:

a) do a binary comparison of the `this.props.appVersions` and `nextProps.appVersions`, but there are other ways that this prop can be altered (sorting).
b) on every `shouldComponentUpdate`, loop through the array and compare the values to the current state. They would have to be mapped to each other by their ID. I'd imagine doing a `.forEach` and a `.find` on every update to compare values would be more costly than the current implementation.
c) What is there currently just seems simpler than b).

ui/src/containers/AppVersionsTableContainer.js
craicoverflow: Oh I think I misunderstood. No I don't believe we need a constructor.

craicoverflow: https://hackernoon.com/the-constructor-is-dead-long-live-the-constructor-c10871bea599

damienomurchu: Cheers, thanks for the link - hadn't realised that, thanks for the link, seems we can do without them! Woohoo - less code :D

ui/src/containers/AppVersionsTableContainer.js
damienomurchu: Yes, agree - had forgotten we were using `appVersions` to store the sort order. This implementation is great :+1:

craicoverflow:
@damienomurchu 

> What I notice when I update the version message field and confirm the save is that the change does not persist.

Good spot. Fixed that now. Nothing to do with the API as the Save button does not connect to the API. 

Adding this fixed it:

```jsx
componentDidUpdate () {
    this.setState({
      canUpdateComponent: false
    });
}
```

> An alternative approach may be to store the message field values in local state, and bind to those instead.

I have tried that :disappointed: With using local state we still need to call `this.setState` which will trigger a re-render.

laurafitzgerald:
> @damienomurchu
> 
> > What I notice when I update the version message field and confirm the save is that the change does not persist.
> 
> Good spot. Fixed that now. Nothing to do with the API as the Save button does not connect to the API.
> 
> Adding this fixed it:
> 
> ```js-jsx
> componentDidUpdate () {
>     this.setState({
>       canUpdateComponent: false
>     });
> }
> ```
> 
> > An alternative approach may be to store the message field values in local state, and bind to those instead.
> 
> I have tried that 😞 With using local state we still need to call `this.setState` which will trigger a re-render.

I also tried lots of combinations of setting a boolean to be used by ShouldComponentUpdate and other React lifecycle methods. This only resulted in the screen not loading the versions on refresh. The user would need to navigate to the landing page and click one of the rows for it to work correctly. Is there a problem with the screen rerending onblur as the main issue was the user would have to select the input box each time a letter was typed. If the user has left the input box (i.e. onblur) wouldn't they expect that they would need to click on it again?

PR #130, opened 2019-04-02
(Aerogear 8913) - Rest API changes to allow GET and POST App 

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8913

## What
- Create POST endpoint to create app
- Impl QueryParameter into GET endpoint to return the app by appID
- Create PATCH endpoint to update app

## Why
- To allow GET and create and update apps using the API
- Attend MDC/Operator requirements

## How
- Create new POST endpoint "/api/apps""
- Add QueryParam "?appId" to the /api/apps GET endpoint

## Verification Steps
- [ ] **Check if CI finished with success**
- [ ] **Create an app from the POST endpoint**
* Call the endpoint `localhost:3000/api/apps` with the following body.
```json
{"appId": "com.aerogear.testpost", "appName": "Test Post" }
```
Screenshot 2019-04-02 at 16 41 56

* Check that the app was created with success
Screenshot 2019-04-02 at 16 41 32

- [ ] **Update the name of an app already created by the POST endpoint** 
* Call the endpoint `localhost:3000/api/apps` with the following body.
```json
{"appId": "com.aerogear.testpost", "appName": "Test Post New Name" }
```
Screenshot 2019-04-02 at 16 42 09

* Check that the app was updated with success
Screenshot 2019-04-02 at 16 42 25

- [ ] **Check to GET the app by AppID**
* Call the the GET endpoint with the Query Param `localhost:3000/api/apps?appId=com.aerogear.testpost`
Screenshot 2019-04-02 at 16 45 03

- [ ] **re-binding the app - Update the deleted_at with NULL when the post is called**
* Set the current data on the field deleted_at of the app

```sql
UPDATE app
SET deleted_at = NOW()
WHERE app_id = 'com.aerogear.testpost'
```

* Call the POST endpoint for the app `localhost:3000/api/apps` with the appId in the body.
Screenshot 2019-04-02 at 16 55 16

* Check that the deleted_at is NULL now. 
Screenshot 2019-04-02 at 16 55 30

## Checklist:

## Progress

- [X] Finished task
- [ ] TODO

## Aditional
* Some changes were requested
* The name update is now in a PATCH endpoint for the apps.

### DIFF
--- a/pkg/web/apps/apps_http_handler.go
+++ b/pkg/web/apps/apps_http_handler.go
@@ -18,7 +18,7 @@ type (
 		DisableAllAppVersionsByAppID(c echo.Context) error
 		HealthCheck(c echo.Context) error
 		DeleteAppById(c echo.Context) error
-		PostApp(c echo.Context) error
+		CreateApp(c echo.Context) error
 	}
 
 	// httpHandler instance
@@ -146,7 +146,7 @@ func (a *httpHandler) DisableAllAppVersionsByAppID(c echo.Context) error {
 
 }
 
-func (a *httpHandler) PostApp(c echo.Context) error {
+func (a *httpHandler) CreateApp(c echo.Context) error {
 
 	// Transform the body request in the version struct
 	app := models.App{}
@@ -157,7 +157,7 @@ func (a *httpHandler) PostApp(c echo.Context) error {
 		return httperrors.BadRequest(c, "Invalid data")
 	}
 
-	err := a.Service.CreateUpdateApp(app)
+	err := a.Service.CreateApp(app)
 
 	if err != nil {
 		return httperrors.GetHTTPResponseFromErr(c, err)
--- a/pkg/web/apps/apps_http_handler.go
+++ b/pkg/web/apps/apps_http_handler.go
@@ -18,6 +18,7 @@ type (
 		DisableAllAppVersionsByAppID(c echo.Context) error
 		HealthCheck(c echo.Context) error
 		DeleteAppById(c echo.Context) error
+		PostApp(c echo.Context) error
 	}
 
 	// httpHandler instance
@@ -119,6 +120,27 @@ func (a *httpHandler) DisableAllAppVersionsByAppID(c echo.Context) error {
 	}
 
 	return c.JSON(http.StatusOK, "")
+
+}
+
+func (a *httpHandler) PostApp(c echo.Context) error {
+
+	// Transform the body request in the version struct
+	app := models.App{}
+	errV := json.NewDecoder(c.Request().Body).Decode(&app)
+
+	// check if the data sent is in the correct format
+	if errV != nil || len(app.AppID) < 1 {
+		return httperrors.BadRequest(c, "Invalid data")
+	}
+
+	err := a.Service.CreateUpdateApp(app)
+
+	if err != nil {
+		return httperrors.GetHTTPResponseFromErr(c, err)
+	}
+
+	return c.NoContent(http.StatusCreated)
 
 }
 
--- a/pkg/web/apps/apps_http_handler.go
+++ b/pkg/web/apps/apps_http_handler.go
@@ -19,6 +19,7 @@ type (
 		HealthCheck(c echo.Context) error
 		DeleteAppById(c echo.Context) error
 		CreateApp(c echo.Context) error
+		UpdateAppNameByID(c echo.Context) error
 	}
 
 	// httpHandler instance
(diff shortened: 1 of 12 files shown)


pkg/web/apps/apps_http_handler.go
```diff
@@ -34,20 +35,43 @@ func NewHTTPHandler(e *echo.Echo, s Service) HTTPHandler {
...
 	// If no apps have been found, return a HTTP Status code of 204 with no response body
-	if err == models.ErrNotFound {
+	if err == models.ErrNotFound && !hasAppId {
 		return c.NoContent(http.StatusNoContent)
 	}
 
+	// If the appId was informed than it should return 404
+	if err == models.ErrNotFound && hasAppId {
+		return c.NoContent(http.StatusNotFound)
+	}
```
wei-lee: it is not a big deal, but I think this change is unnecessary to return a different status code when there is an app in in the request. It gives more information away about the system for no obvious benefit. In the worst scenario, it means someone could use this information to guess the appid in the system. It's probably doesn't matter that much for this system, and it actually makes the code more complicated. Given that this wasn't there, I would suggest keep it the old way.

camilamacedo86: @wei-lee all project is returning 404 when the app is not found. For the app is not found it does not exist and/or is with the deleted_at <> NULL. I am just following how it was defined to get done.

davidffrench: I would agree with Wei on this, it makes sense to leave it as just

```
if err == models.ErrNotFound {
	return c.NoContent(http.StatusNoContent)
}
```

pkg/web/apps/apps_service.go
```diff
@@ -108,23 +121,37 @@ func (a *appsService) UnbindingAppByAppID(appID string) error {
...
+	// check if is disabled
+	if appStored.DeletedAt != "" {
+		// if is deleted so just reactive the existent app
+		if err := a.repository.UnDeleteAppByAppID(app.AppID); err != nil {
+			return err
+		}
+	}
+
+	// update the name if it was changed
+	if app.AppName != "" && appStored.AppName != app.AppName {
```
davidffrench: We should not be updating the appName on a /apps POST. If a user wants to update a field within a specific resource it would be a PATCH on `/apps/:id`.

pkg/web/apps/apps_http_handler.go
camilamacedo86: @davidffrench and all other places that are returning 404 when the app is not found and or have the deleted_at <> NULL?

pkg/web/apps/apps_service.go
camilamacedo86: OK. But the disabled_at is also updated here. Should we create an endpoint for update the name only using the PATCH method or since the deleted_at update is done here we can move with it in the post as well?

pkg/web/apps/apps_http_handler.go
davidffrench: I don't mind if it is a 204 or 404. I think 404 is confusing in the context of an empty REST collection.

The main point of Weis comment was that there is a different response status code sent for when a model is not found and the appId filter is sent as a query param and when it is not. This should not be the case, the response code should be consistent for this single endpoint when no resources were found.

pkg/web/apps/apps_service.go
davidffrench: The deleted_at is an internal only field, it is never sent from an end user to be updated. It captures the soft deletion of this entity, this is why we are ok to update the internal deleted_at field on this POST to undo the soft deletion.

An update to an individual resource from an end user should be done through a PUT or PATCH of the individual resource rather than through a filter on the resources collection.

camilamacedo86: Question for @wei-lee ^ Will be possible change the appName from MDC?

camilamacedo86: Hi @davidffrench,

I don't know but I think that yes. So, I will move this for a PACTH if is possible  change the name by MDC. @wei-lee could you let us know if users will or not be able to change the name of the app by the MDC?

wei-lee: yes, it's possible in MDC. however, I don't think we are going to update the app names across all services at least for now. It is a bit tricky to do and we probably need to think about it a bit more.

pkg/web/router/router.go
```diff
@@ -152,6 +161,53 @@ func SetAppRoutes(r *echo.Group, appsHandler apps.HTTPHandler) {
...
+	//   description:
+	//   required: true
+	//   schema:
+	//     $ref: '#/definitions/App'
+	// responses:
+	//   201:
+	//     description: successful operation
+	//   400:
+	//     description: Invalid data supplied
+	r.PATCH("/apps/:id", appsHandler.UpdateAppNameByID)
```
davidffrench: A PATCH on a resource should technically allow the user to update any field for an app resource, not just the single app name field. For now this will be ok. We could make it more generic in the future.

camilamacedo86: But, if it should allow updating any field then shows that it would should be a PUT and not a PATCH.

davidffrench: A PUT would be updating the entire resource. PATCH allow you to update 1-many fields within a resource.

This implementation of PATCH only allows the user to update the app name. It is not generic so that if we added new fields to the resource that it would work automatically.

camilamacedo86: We don't have other fields by the business roles that we can update anyway. Just the name and deleted_at. However, I understand that the deleted_at should not be updated here too.

PR #137, opened 2019-04-05
feat(ui): Add react hooks eslint rules

davidffrench:
## What
Adding react hooks eslint plugin and rules

## Why
To catch errors and warning for our react hooks

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/ui/.eslintrc
+++ b/ui/.eslintrc
@@ -1,7 +1,10 @@
 {
     "extends": ["react-app","standard"],
+    "plugins": ["react-hooks"],
     "rules": {
         "semi": [2, "always"],
-        "prefer-const": [2]
+        "prefer-const": [2],
+        "react-hooks/rules-of-hooks": "error",
+        "react-hooks/exhaustive-deps": "warn"
     }
 }
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -6012,6 +6012,12 @@
         "prop-types": "^15.6.2"
       }
     },
+    "eslint-plugin-react-hooks": {
+      "version": "1.6.0",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-react-hooks/-/eslint-plugin-react-hooks-1.6.0.tgz",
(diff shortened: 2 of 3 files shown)


ui/.eslintrc
```diff
@@ -1,7 +1,10 @@
 {
     "extends": ["react-app","standard"],
+    "plugins": ["react-hooks"],
```
craicoverflow: Should we also be adding the Patternfly React plugin?

ui/.eslintrc
```diff
@@ -1,7 +1,10 @@
...
     "rules": {
         "semi": [2, "always"],
-        "prefer-const": [2]
+        "prefer-const": [2],
+        "react-hooks/rules-of-hooks": "error",
+        "react-hooks/exhaustive-deps": "warn"
```
craicoverflow: Would there be an additional benefit of setting this level to "error"?

There are warnings in the code already, so we may miss them now because they are done. With it set to "error" we can ensure we find and fix them.

craicoverflow: T

craicoverflow:
Verified with the following code change:

```js
if (apps.length) {
    useEffect(getApps(), []);
}
```

ui/.eslintrc
davidffrench: These are the recommendations from react

ui/.eslintrc
davidffrench: separate PR for that. I didn't want to get bogged down with the patternfly eslint stuff

craicoverflow:
> Looks good. CI failing because UI refactored tests haven't been merged yet, so suggest not merging this until that PR is merged first.

Well we would be merging into his feature branch, which is broken anyway. This PR isn't what is causing the failed tests. We will have to merge our other PRs into the refactor branch too, before the CI is passing.

davidffrench:
Getting back to this now. I agree with Enda. We will have to start merging things to this feature branch.

PR #138, opened 2019-04-05
fix(ui): Refactor reducer tests

davidffrench:
## What
Creating reducer tests after the refactor

## Checklist:

## Progress

- [ ] Finished task
- [ ] TODO


ui/src/actions/actions-ui.js
```diff
@@ -12,7 +12,7 @@ export const appsTableSort = (_event, index, direction) => {
   };
 };
 
-export const appDetailsSort = (_event, index, direction) => {
+export const appVersionsTableSort = (_event, index, direction) => {
```
damienomurchu: does `_event` need to be passed in here?

ui/src/reducers/app.js
```diff
@@ -33,35 +33,40 @@ export const cloneAppData = (appData) => {
 
 export default (state = initialState, action) => {
   switch (action.type) {
-    case APP_VERSIONS_SORT:
+    case APP_VERSIONS_SORT: {
```
damienomurchu: Good catch & fix in adding a block scope for each switch case implementation :+1:

ui/src/selectors/index.js
```diff
@@ -44,15 +49,26 @@ const areColumnValuesEqual = (rows, index) => {
...
 
 /**
  * Selector to convert app objects to array of app values
- *
- * @param {Array} apps - Array of app objects
+ * @param {Object} state Redux state
+ * @param {Object} _sortBy sortBy includes index and direction
  */
-export const getAppsTableRows = (state) => {
+export const getAppsTableRows = (state, _sortBy) => {
```
damienomurchu: Just curious - what rule(s) are we following to decide if a variable should be prefixed with `_` ?

ui/src/actions/actions-ui.js
craicoverflow: I would imagine they do need to be @damienomurchu - at least with how it is currently implemented.

The action is being called directly from the Patternfly table, and `_event` is an argument that comes from the table. So if it were to be removed, `index` would become `_event`.

https://github.com/aerogear/mobile-security-service/blob/686a49a8edb8868555d37477145ea1eeef5abcbb/ui/src/containers/appView/AppVersionsTableContainer.js#L85

davidffrench: The alternative is to have another function and only pass in the index and direction.

ui/src/reducers/app.js
davidffrench: This came from @craicoverflow . Thumbs up to him

ui/src/selectors/index.js
davidffrench: If it is not used, it is generally prefixed with an `_`

ui/src/actions/actions-ui.js
damienomurchu: yes, that was what I was thinking @davidffrench, as at the moment the reducer is being called directly instead of via a handler

ui/src/selectors/index.js
damienomurchu: @davidffrench got ya, so is `_sortBy` required at all as a parameter in this method?

davidffrench: it is not required. I thought it would be good to list it though to show that all selector inputs get the same parameters

ui/src/actions/actions-ui.js
davidffrench: Just less code to call the action creators directly rather than have an intermediate function.

ui/src/actions/actions-ui.test.js
```diff
@@ -0,0 +1,86 @@
+import * as types from './types';
+import * as actions from './actions-ui';
+
+describe('actions', () => {
+  it('appsTableSort should create APPS_SORT action', () => {
+    const index = 0;
+    const direction = 'ASC';
```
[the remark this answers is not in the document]
davidffrench: done

ui/src/actions/actions-ui.test.js
```diff
@@ -0,0 +1,86 @@
...
+      payload: {
+        index: index,
+        direction: direction
+      }
+    });
+  });
+
+  it('appVersionsTableSort should create APP_VERSIONS_SORT action', () => {
+    const index = 1;
+    const direction = 'DESC';
```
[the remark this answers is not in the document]
davidffrench: done

PR #141, opened 2019-04-08
feat(ui): Get username from the backend and populate - else logout (A…

laurafitzgerald:
…EROGEAR-9032)

## Notes
Creating as a draft pr as it the backend work captured in [AEROGEAR-9039](https://issues.jboss.org/browse/AEROGEAR-9039). **Please do not merge this pr until that work is completed as it blocks the ui from loading**
 

## Motivation
JIRA: https://issues.jboss.org/browse/AEROGEAR-9032 

## What
Use the user data if it's returned
On getting the user data is failed, then the user should be redirected to the oauth sign in screen

## Verification Steps
1. Build the docker image locally
`docker build -t docker.io/aerogear/mobile-security-service:AEROGEAR-9032-verification .`

2. Push Image to dockerhub
`docker push docker.io/aerogear/mobile-security-service:AEROGEAR-9032-verification`

3. Deploy the image to the project in the community cluster
3a. Contact me for link to project
3b. Update the deployment config `image` to the new docker image
3c. Wait for the deployment to finish and the pod to come up.

4. Navigate to the mss route

5. You should be presented with the OAuth login screen. 

6. Log in

EDIT: The following step is no longer required as that will be implemented seperately.
7. As the backend logic for the user functionality has not yet been implemented the user will be redirected to the the log in screen.

## Checklist:

## Progress

- [x] UI implementation
- [x] Backend implementation


ui/src/containers/HeaderContainer.js
```diff
@@ -19,30 +24,36 @@ const HeaderContainer = ({ currentUser, history }) => {
...
-    <Header
-      currentUser={currentUser}
-      appName={config.app.name.toUpperCase()}
-      isDropDownOpen={isDropDownOpen}
-      onUserDropdownToggle={onUserDropdownToggle}
-      onTitleClick={onTitleClick}
-      logout={logout}
-    />
-  );
+  if (isUserRequestFailed) {
```
laurafitzgerald: I think this might solve the issue we saw on MDC where the screen loads with no data if the URL is manipulated after logout. wdyt @davidffrench @wei-lee

ui/src/containers/HeaderContainer.js
```diff
@@ -19,30 +24,36 @@ const HeaderContainer = ({ currentUser, history }) => {
...
+    logout();
```
laurafitzgerald: I think this might solve the issue we saw on MDC where the screen loads with no data if the URL is manipulated after logout. wdyt @davidffrench @wei-lee

davidffrench: Yes it likely would but there are side effects. The user could be legitimately logged in but this specific request to the server fails for whatever reason, we will then be logging the user out.

laurafitzgerald: Yes but if requests to the api are failing then the application wont work, and I would consider that a bug. I agree there would be no feedback to the user as to why they are getting logged out. Is that something we should put in the logs of the container so it could be debugged?

davidffrench: We do get a 401 response back from this endpoint on authentication failure. With this solution, we should be specifically checking the HTTP response status code for all requests and if it is a 401, display a toast to the user with a message and link advising them to login again.

This will also take into account the token expiring if the page is left open for a long period of time and does not automatically redirect them without them necessarily knowing why.

laurafitzgerald: @davidffrench you make a good point, it does sound like a separate piece of work that we could discuss - handling non 200 response codes - we could discuss it at refinement tomorrow?

laurafitzgerald: after some discussion, removing this check for now. There is a seperate task to deal with NotFounds here: https://issues.jboss.org/browse/AEROGEAR-9049 so resolving this comment.

laurafitzgerald:
@davidffrench @craicoverflow I've taken this pr out of draft and removed blocked and draft status. The offending line has been removed and will be dealt with separately. This can now be verified and merged once approved, could you guys take a look when you have time.
Note: I didn't do any tests here as I understand they are being refactored now. Let me know if I should create the tests here.

ui/src/reducers/header.js
```diff
@@ -1,9 +1,30 @@
+import {
+  USER_REQUEST,
+  USER_SUCCESS,
+  USER_FAILURE
+} from '../actions/types.js';
+
 const initialState = {
-  currentUser: 'currentUser'
+  currentUser: '',
```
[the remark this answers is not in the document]
laurafitzgerald: We could make it a user object which contains the username and mail as both should be returned from the api. wdyt?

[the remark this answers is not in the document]
craicoverflow: Yeah I think it should be. It will also make it more extensible in the future in case we need to add more user related state.

ui/src/containers/HeaderContainer.js
```diff
@@ -1,13 +1,18 @@
...
 import Header from '../components/Header';
+import { getUser } from '../actions/actions-ui';
 
-const HeaderContainer = ({ currentUser, history }) => {
+const HeaderContainer = ({ currentUser, isUserRequestFailed, history, getUser }) => {
   const [isDropDownOpen, setIsDropDownOpen] = useState(false);
 
+  useEffect(() => {
+    getUser();
+  }, []);
```
craicoverflow: eslint gives the following warning:

> React Hook useEffect has a missing dependency: 'getUser'. Either include it or remove the dependency array. If 'getUser' changes too often, find the parent component that defines it and wrap that definition in useCallback.eslint(react-hooks/exhaustive-deps)

ui/src/reducers/header.js
```diff
@@ -1,9 +1,33 @@
...
+  user: {
+    username: '',
+    email: ''
+  },
+  isUserRequestFailed: false
 };
 
 export default (state = initialState, action) => {
   switch (action.type) {
+    case USER_REQUEST:
```
craicoverflow: I suggest wrapping each case in a block, like this:

https://github.com/aerogear/mobile-security-service/blob/c0ca7cba530092ced9be67a9ed18fea150561ef2/ui/src/reducers/modals.js#L39

This allows variables to only exist in that case, allowing for reuse of variable names. We are in the process of adding this to the other reducers at the minute.

ui/src/containers/HeaderContainer.js
laurafitzgerald: @craicoverflow this one will be resoved as part of this JIRA: https://issues.jboss.org/browse/AEROGEAR-9050 Resolving for now.

craicoverflow:
Verified that this shows the username in the header.

laurafitzgerald:
@craicoverflow I made the changes as requested and added some additional tests.(Thanks @davidffrench) Please rereview when you get a chance and feel free to merge this one yourself. It is rebased on the Refactor-UI branch at time of writing.

PR #142, opened 2019-04-09
feat(auth): adding user endpoint to auth against oauth (AEROGEAR-9033)

laurafitzgerald:
## Motivation
JIRA: https://issues.jboss.org/browse/AEROGEAR-9033

## What
Add a endpoint to GET the user which can be used to validate if the user has authenticated with oauth in the UI

## How
Added endpoint to swagger
Added a http_handler for the user resource
Added a service layer for the user resource
Added a model for the user resource
Added the route to the router

## Verification Steps
 
1. Run make serve
2. Send a GET request to the user endpoint with the following headers
 2.a X-Forwarded-User
 2.b X-Forwarded-Email
3. The returned request should be in the following format but with the values you passed

Screenshot 2019-04-09 at 09 51 01

## Checklist:

## Progress

- [x] Functionality
- [x] Tests
- [ ] Verify against the [UI pr ](https://github.com/aerogear/mobile-security-service/pull/141)

### DIFF
--- a/cmd/mobile-security-service/main.go
+++ b/cmd/mobile-security-service/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aerogear/mobile-security-service/pkg/web/checks"
 	"github.com/aerogear/mobile-security-service/pkg/web/initclient"
 	"github.com/aerogear/mobile-security-service/pkg/web/router"
+	"github.com/aerogear/mobile-security-service/pkg/web/user"
 	dotenv "github.com/joho/godotenv"
 	"github.com/labstack/echo"
 	log "github.com/sirupsen/logrus"
@@ -94,6 +95,12 @@ func setupServer(e *echo.Echo, c config.Config, dbConn *sql.DB) {
 	// Prefix api routes
 	APIRoutePrefix := c.APIRoutePrefix
 	apiGroup := e.Group(APIRoutePrefix)
+
+	// User handler setup
+	userHandler := user.NewHTTPHandler(e)
+
+	// Setup user routes
+	router.SetUserRoutes(apiGroup, userHandler)
 
 	// App handler setup
 	appsPostgreSQLRepository := apps.NewPostgreSQLRepository(dbConn)
--- a/pkg/web/user/user_http_handler.go
+++ b/pkg/web/user/user_http_handler.go
@@ -1,0 +1,47 @@
+package user
+
+import (
+	"net/http"
+
+	"github.com/aerogear/mobile-security-service/pkg/httperrors"
+	"github.com/aerogear/mobile-security-service/pkg/models"
+	"github.com/labstack/echo"
+)
+
+const (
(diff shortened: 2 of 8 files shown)


laurafitzgerald:
@davidffrench @craicoverflow thanks for the reviews. I will address the comments in a follow on commit. I intentionally didn't add tests initially in order to wait for the review. I will add those now. :)

laurafitzgerald:
@davidffrench @craicoverflow changes made as requested. Tests are passing locally. Could you take a look at this again when you have some time. Thanks :)

cmd/mobile-security-service/main.go
```diff
@@ -95,6 +96,13 @@ func setupServer(e *echo.Echo, c config.Config, dbConn *sql.DB) {
 	APIRoutePrefix := c.APIRoutePrefix
 	apiGroup := e.Group(APIRoutePrefix)
 
+	// User handler setup
+	userService := user.NewService()
```
davidffrench: this is no longer required now.

pkg/web/user/user_http_handler.go
```diff
@@ -0,0 +1,47 @@
...
+// GetUser returns the user if Oauth added headers are present
+func (a *httpHandler) GetUser(c echo.Context) error {
+
+	var user models.User
+
+	//these headers values will be set by the openshift oauth-proxy
+	if userNameHeader := c.Request().Header[USER_NAME_HEADER]; userNameHeader != nil && userNameHeader[0] != "" {
+		user.Username = userNameHeader[0]
+	} else {
+		return httperrors.NotFound(c, "No User Found")
```
craicoverflow: Should this return a 401? Where does a 401 get returned to the client?

pkg/web/user/user_http_handler_test.go
```diff
@@ -0,0 +1,50 @@
...
+			wantErr:  true,
+			wantCode: 404,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			//Setup
+			e := echo.New()
+			req := httptest.NewRequest(http.MethodGet, "/", nil)
+			if !tt.wantErr {
```
craicoverflow: I would put these in the test table above. Something like:

```go
{
    name:     "Should return success when header provided",
    wantErr:  false,
    wantCode: 200,
    headers: &headers{
        username: "TestUser",
        email: "test@user.com"
    }
},

...
...

req.Header.Add(USER_NAME_HEADER, tt.headers.username)
req.Header.Add(USER_EMAIL_HEADER, tt.headers.email)
```

pkg/web/user/user_http_handler.go
laurafitzgerald: oauth returns a 401 as described here: https://github.com/openshift/oauth-proxy#endpoint-documentation so we don't need to do that here.

PR #145, opened 2019-04-10
feat(app) : Add new for apps struct (AEROGEAR-8914)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8914

## What
Create fun to create the object

## Why
Shows a good practice use the MSS model instead of replicated it in the operator project

## How
Just add the new func for we are able to create the struct

### DIFF
--- a/pkg/models/app.go
+++ b/pkg/models/app.go
@@ -12,3 +12,13 @@ type App struct {
 	DeployedVersions      *[]Version `json:"deployedVersions,omitempty"`
 	DeletedAt             string     `json:"deletedAt,omitempty"`
 }
+
+// NewAppByNameAndAppID will create a new App object based on the name and appId which indeed are the only values
+// according to the business that can be used/provided in order to the service create an app. This will be used in the
+// systems which would consume the REST API to create new apps. ( E.g Operator )
+func NewAppByNameAndAppID(name, appId string) *App {
+	app := new(App)
+	app.AppName = name
+	app.AppID = appId
+	return app
+}


pkg/models/app.go
```diff
@@ -12,3 +12,10 @@ type App struct {
 	DeployedVersions      *[]Version `json:"deployedVersions,omitempty"`
 	DeletedAt             string     `json:"deletedAt,omitempty"`
 }
+
+func NewAppByNameAppID(name, appId string) *App {
```
craicoverflow: Public functions in go should be commented, with the following format:

```go
// NewAppByNameAppID is a ....
```

craicoverflow: This function is not being used anywhere, why is it being added?

craicoverflow: Suggestion:

```go
// NewApp returns a new instance of App
func NewApp(name string, appID string, deployedVersions *[]Version) *App {
	return &App{
		ID:               uuid.New().String(),
		AppName:          name,
		AppID:            appID,
		DeployedVersions: deployedVersions,
	}
}
```

No need for a specific constructor by app ID and app name in my opinion You can just add one extra parameter and pass nil. Also, it is more common to use composite literals over `new`. This is suggested in the official docs: [composite literals](https://golang.org/doc/effective_go.html#composite_literals).

camilamacedo86: Hi @craicoverflow,

According to the business no make sense pass the other attributes. The only information that should be used to create the app is the name and the appId. The UUID should be created by the Server when it will be added. 

Also, note that one class can have many constructors as it is required and because of this and because of the business roles of this object I do not agree with your suggestion.

camilamacedo86: It will be used by the operator.

camilamacedo86: I will add the comment.

craicoverflow: function should be called `NewAppByNameAndAppID`

craicoverflow: If all that is needed to create an app is name and appID, you could just call it `NewApp`.

craicoverflow: Okay, well I still think you should return a composite literal.

camilamacedo86: It is returning a composite literal. "We can simplify it using a composite literal, which is an expression that creates a new instance each time it is evaluated."

It will create a new instance each time it is evaluated based on the only info that can be provided to do it. The UUID should not be created in this case. I don't want to add the UUID and the UUID is created when the POST REST is called in its respective service implementation.

craicoverflow: Please read the documentation I sourced. It literally explains the difference between a composite literal:

> There's a lot of boiler plate in there. We can simplify it using a composite literal, which is an expression that creates a new instance each time it is evaluated. I am not talking about the UUID, that is fine.

```go
func NewFile(fd int, name string) *File {
    if fd < 0 {
        return nil
    }
    f := File{fd, name, nil, 0}
    return &f
}
```

PR #151, opened 2019-04-12
fix(apps-view): fixed issue where page render failed when no apps were returned from server

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9060

## What

There was an error where the apps view failed to render if no apps were returned from the server. The UI was attempting to perform a `map` on `undefined` which caused the error,

## Why

Page failed to load.

## How

A 204 response is returned when no apps exist in the database. For this reason the APPS_FAILURE action was never hit as a 204 is a successful response. I added an additional check in the `DataService` module to return an empty array instead of undefined. This solved the issue.

## Verification Steps

1. Ensure that the database has 0 apps.
2. Open the UI.
3. You should see `Unable to fetch any apps :/`.
4. Verify tests are passing.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

I also fixed some warnings in the browser I noticed that aren't directly related to this PR.

### DIFF
--- a/ui/src/DataService.js
+++ b/ui/src/DataService.js
@@ -43,6 +43,11 @@ const request = async (path, method, body) => {
  */
 const fetchItems = async url => {
   const result = await request(url, 'GET');
+  // if a 204, return an empty array
+  if (result === true) {
+    return [];
+  }
+
   return result || [];
 };
 
--- a/ui/src/components/appView/AppOverview.js
+++ b/ui/src/components/appView/AppOverview.js
@@ -27,7 +27,7 @@ const AppOverview = ({ className, app }) => {
 AppOverview.propTypes = {
   className: PropTypes.string.isRequired,
   app: PropTypes.shape({
-    appId: PropTypes.string.isRequired
+    appId: PropTypes.string
   }).isRequired
 };
 
--- a/ui/src/containers/AppsTableContainer.js
+++ b/ui/src/containers/AppsTableContainer.js
@@ -14,7 +14,6 @@ import { getSortedAppsTableRows } from '../selectors/index';
  * @param {array} props.apps Array of app object to display
  * @param {array} props.appRows Transformed app array of arrays contains just the values needed for the table
  * @param {object} props.sortBy Contains column index and direction for sorting
- * @param {boolean} props.isAppsRequestFailed Boolean on if the apps request has failed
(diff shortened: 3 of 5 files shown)


ui/src/DataService.js
```diff
@@ -43,6 +43,11 @@ const request = async (path, method, body) => {
  */
 const fetchItems = async url => {
   const result = await request(url, 'GET');
+  // if a 204, return an empty array
+  if (result === true) {
```
damienomurchu: should this not be if result is false?

ui/src/components/appView/AppOverview.js
```diff
@@ -27,7 +27,7 @@ const AppOverview = ({ className, app }) => {
 AppOverview.propTypes = {
   className: PropTypes.string.isRequired,
   app: PropTypes.shape({
-    appId: PropTypes.string.isRequired
+    appId: PropTypes.string
```
damienomurchu: good spot :+1:

ui/src/DataService.js
craicoverflow: Look at the `request` function - it is returning true if the response is 204.

ui/src/containers/AppsTableContainer.js
```diff
@@ -40,7 +38,7 @@ export const AppsTableContainer = ({
 
   useEffect(() => {
     getApps();
-  }, []);
+  }, [getApps]);
```
damienomurchu: I saw this before as a missing dependency - are there any side-effects of adding it here do you know?

craicoverflow: Here is the warning message:

```
React Hook useEffect has a missing dependency: 'getApps'. Either include it or remove the dependency array.
```

It seems to indicate that an empty array is not the right thing to do. It doesn't seem to have any side effect.

ui/src/components/appView/AppOverview.js
craicoverflow: Yeah, if the `app` is required then this will automatically be required.

ui/src/DataService.js
damienomurchu: Ok, understand now :+1:

damienomurchu:
Code changes look good @craicoverflow. 

Haven't verified them, and just going to do that now

PR #154, opened 2019-04-17
feat(auth): added a redirect before loading routes if not logged in (AEROGEAR-9072) (#154)feat(auth): added a redirect before loading routes if not logged in (AEROGEAR-9072) (#154)feat(auth): added a redirect before loading routes if not logged in (AEROGEAR-9072) (#154)Merge pull request #154 from craicoverflow/AEROGEAR-9072

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9072

## What

Added an automatic redirect to the login screen if a user is not logged in.

## Why

This prevents anonymous users from gaining access to the UI. Even though the data would not return, it is still best practice.

## How
 
Converted `App.js` from a stateless component to a stateful one, connected the Redux store. Added a check in the `App.js` component and redirect to login screen when the user check fails.

## Verification Steps

1. Open the community cluster instance. Contact me for the URL.
2. Ensure that the image name for the `mobile-security-service-app` is `ephelan/mobile-security-service`.
3. Login to the security service.
4. Log out and try to access the home page. You should be redirected to the login screen.

## Checklist:

## Progress

- [ ] Finished task
- [ ] Complete test


ui/src/components/App.js
```diff
@@ -1,12 +1,36 @@
...
 /**
- * Main entry point for the entire UI view
+ * Main entry point for the entire UI view.
+ *
+ * @param {func} authenticate - Redux action to check if the user is logged in.
+ * @param {bool} isLoggedIn - Whether the user is logged in or not.
+ * @param {bool} isLoading - Whether the component has finished loading.
  */
-const App = () => {
+const App = ({ authenticate, isLoggedIn, isLoading }) => {
```
damienomurchu:
```
export const App = ({ authenticate, isLoggedIn, isLoading }) => {
```
You'll need to export the App var to be able to test it without Redux

ui/src/components/App.spec.js
```diff
@@ -1,11 +1,14 @@
 import React from 'react';
 import { createRenderer } from 'react-test-renderer/shallow';
-import { BrowserRouter as Router } from 'react-router-dom';
 import App from './App';
```
damienomurchu:
```suggestion
import { App } from './App';
```
This will import just the App component without Redux (after the above change to export `const App` is made in `App.js`

ui/src/components/App.spec.js
```diff
@@ -19,11 +22,16 @@ const setup = (propOverrides) => {
...
+ */
 describe('components', () => {
   describe('App', () => {
     it('should render', () => {
-      const { output } = setup();
-      expect(output.type).toBe(Router);
+      const { output } = setup({ isLoggedIn: true, isLoading: true });
+
+      const app = output.props.children;
+      expect(app.type).toBe(undefined);
```
damienomurchu:
```suggestion
      expect(app.type).toBe(Router);
```
Once you export `App` without router & import it above as suggested, you should be able to test the component renders a `Router` object as per the render method

ui/src/components/App.spec.js
```diff
@@ -19,11 +22,16 @@ const setup = (propOverrides) => {
...
+ * @todo get these tests working
+ */
 describe('components', () => {
   describe('App', () => {
     it('should render', () => {
-      const { output } = setup();
-      expect(output.type).toBe(Router);
+      const { output } = setup({ isLoggedIn: true, isLoading: true });
+
+      const app = output.props.children;
```
damienomurchu:
```suggestion

```
This line should no longer be necessary, as you do not need to drill down into the `output` object, as the `output` object should just render a `Router` object as per the render method, which you can test for as was done previously.

ui/src/containers/HeaderContainer.js
```diff
@@ -56,8 +56,8 @@ HeaderContainer.propTypes = {
 
 function mapStateToProps (state) {
   return {
-    currentUser: state.header.user,
-    isUserRequestFailed: state.header.isUserRequestFailed
+    currentUser: state.user.user,
```
damienomurchu:
```suggestion
    currentUser: state.user.username,
```
Should this be user.username?

ui/src/reducers/user.js
```diff
@@ -1,15 +1,19 @@
...
   USER_SUCCESS,
-  USER_FAILURE
+  USER_FAILURE,
+  USER_AUTHENTICATE_SUCCESS,
+  USER_AUTHENTICATE_FAILURE
 } from '../actions/types.js';
 
 const initialState = {
   user: {
     username: '',
```
damienomurchu: Know it's not part of this PR, but:

```suggestion
    name: '',
```

damienomurchu:
Seeing as we're now connecting `App` to redux, I think we should move it from `/components` to `/containers`

ui/src/AuthService.js
```diff
@@ -0,0 +1,55 @@
...
+  return response;
+};
+
+/**
+ * Makes a HTTP GET request to a provided URI and returns a single item
+ *
+ * @param {string} url - The full URL to make the GET request to
+ */
+const fetchItem = async url => {
+  const result = await request(url, 'GET');
```
laurafitzgerald: The value passed here is url but it's accepted as path in the async method. That method also has a url value which is coming from the the config. It could be bit clearer to call url path here?

ui/src/AuthService.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Creates a configuration obect with information about the request type
+ * and the request body
+ *
+ * @param {string} method - HTTP Method (PUT, GET, PATCH, DELETE, POST)
+ * @param {object|Array} body - JSON Payload to send in the request
+ * @returns a config object to send in HTTP requests
+ */
+const requestConfig = (method, body) => ({
```
laurafitzgerald: Out of interest what is the motivation behind having a separate service for auth? Is it because it's calling to oauth rather than to our backend service?

ui/.dockerignore
```diff
@@ -0,0 +1 @@
+.env
```
laurafitzgerald: 👍

ui/src/containers/HeaderContainer.js
laurafitzgerald: Let's remove the outer user object in the reducer so we don't have the nested objects here. We're doing that in the other reducer states.

ui/src/reducers/user.js
```diff
@@ -1,15 +1,19 @@
...
     email: ''
   },
-  isUserRequestFailed: false
+  isUserRequestFailed: false,
+  isLoggedIn: false,
+  isLoading: true
```
laurafitzgerald: What is the isLoading value used for?

ui/src/reducers/user.js
laurafitzgerald: @damienomurchu it's the username we get back from oauth. The user's name doesn't get sent so username makes mores sense here.

ui/src/AuthService.js
craicoverflow: That's true, I just copied what's in the DataService. I will update it.

ui/src/AuthService.js
craicoverflow: Yes because it is calling a completely separate service. The DataService is coupled to the API.

ui/src/containers/HeaderContainer.js
craicoverflow: Like what? user.currentUser?

ui/src/reducers/user.js
craicoverflow: See [this line](https://github.com/aerogear/mobile-security-service/pull/154/files#diff-b783c7f57868724b8b576eb3f530dfb1R27).

Without a separate `isLoading` state, all we would have to work with is `isLoggedIn`. `isLoggedIn` is a boolean which starts off as false. As it starts off as false, the page would redirect straight away because it would see `isLoggedIn === false` and do the redirect. 

`isLoading` prevents this. Only when the application knows that a) we have checked if the user is logged in (setting `isLoading` to false) and b) the user is not logged in (setting `isLoggedIn` to false) will a redirect happen.

ui/src/reducers/user.js
damienomurchu: @laurafitzgerald that's a fair point :+1:

docker-compose.yml
```diff
@@ -16,6 +16,8 @@ services:
       PGUSER: postgresql
       PGPASSWORD: postgres
       PGDATABASE: mobile_security_service
+      REACT_APP_API_URL: /api
+      NODE_ENV: development
```
damienomurchu: :+1:

ui/src/containers/HeaderContainer.js
craicoverflow: `user` is the name of the reducer. the second `user` is the user object, containing the email and username. It is a bit confusing, I considered flattening the `user` reducer so that they are all on the same level but thought it better to keep the actual user as a lone object, so it can be imported into components without everything else that is irrelevant.

ui/src/AuthService.js
craicoverflow: > Code looks good @craicoverflow - good work 
> 
> I haven't verified as @laurafitzgerald has already done so.
> 
> Only comment I would make is with 2 services now using retrieving data with fetch maybe we could abstract it out to a seperate http layer, but that may be unnecessary at this point.

I agree, though the `request` function in each are slightly different as the structure of the data returned from each are different. Our API returns a JSON object on error whereas oauth returns a string.

craicoverflow:
@damienomurchu @laurafitzgerald thanks for the reviews. Made some changes and re-requested your reviews.

ui/src/AuthService.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param {string} path - The URI to send the request to
+ * @param {*} method - HTTP Method (PUT, GET, PATCH, DELETE, POST)
+ * @param {*} body - The request payload
+ * @returns {boolean|JSON} - Returns the response from the HTTP request
+ */
+const request = async (path, method, body) => {
+  const response = await fetch(`${config.auth.url}/${path}`, requestConfig(method, body));
+  if (!response.ok) {
+    const msg = await response.text();
```
damienomurchu: I wonder if the `await` is actually needed here, as `response` should be returned from the fetch call at this point - is `.wait()` asynchronous?

ui/src/actions/actions-ui.js
```diff
@@ -82,6 +83,8 @@ export const setAllAppVersionsDisabled = (customMessage) => {
   };
 };
 
+export const authenticate = fetchAction([actions.USER_AUTHENTICATE_SUCCESS, actions.USER_AUTHENTICATE_SUCCESS, actions.USER_AUTHENTICATE_FAILURE], AuthService.authenticate);
```
damienomurchu:
```suggestion
export const authenticate = fetchAction([actions.USER_AUTHENTICATE_REQUEST, actions.USER_AUTHENTICATE_SUCCESS, actions.USER_AUTHENTICATE_FAILURE], AuthService.authenticate);
```

ui/src/containers/App.js
```diff
@@ -0,0 +1,63 @@
...
+    authenticate();
+  }, [authenticate]);
+
+  /**
+   * Check to see if the user is logged in.
+   * If not, redirect them to the login screen.
+   */
+  const checkIfAuthenticated = () => {
+    if (config.mode === 'production' && !isLoading && !isLoggedIn) {
+      window.location.replace(config.auth.signInUrl);
```
damienomurchu: I'm wondering if we should use `window.location.assign` here instead? Effect is the same, but using `replace` we lose the ability to use the back button on the browser

ui/src/containers/HeaderContainer.js
```diff
@@ -30,7 +30,7 @@ export const HeaderContainer = ({ currentUser, history, getUser }) => {
   };
 
   const logout = () => {
-    window.location.replace('/oauth/sign_in');
+    window.location.replace(config.auth.signInUrl);
```
damienomurchu: same as comment above re `replace` and `assign`

ui/src/containers/App.js
craicoverflow: I will try it out. I tried [`react-router/Redirect`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md) before and that did not work.

craicoverflow: I have tried this out and can confirm it will not work as we want. The problem is that `window.location.assign` will allow a user to use the back button to get back to the home page, even if they are not logged in.

craicoverflow: And if they are not logged in it should not matter because they should not be able to use the back button to navigate to it.

laurafitzgerald: Yes agreed, we should use the solution which blocks the user from being able to use the back button which would show them the UI without data.

ui/src/AuthService.js
craicoverflow: Yes it is needed. `fetch` is a web API that returns a `Promise`.



The Fetch API provides an interface for fetching resources (including across the network). It will seem familiar to anyone who has used XMLHttpRequest, but the new API provides a more powerful and flexible feature set.
Concepts and usage
Section

Fetch provides a generic definition of Request and Response objects (and other things involved with network requests). This will allow them to be used wherever they are needed in the future, whether it’s for service workers, Cache API and other similar things that handle or modify requests and responses, or any kind of use case that might require you to generate your own responses programmatically.

It also provides a definition for related concepts such as CORS and the HTTP origin header semantics, supplanting their separate definitions elsewhere.

For making a request and fetching a resource, use the WindowOrWorkerGlobalScope.fetch() method. It is implemented in multiple interfaces, specifically Window and WorkerGlobalScope. This makes it available in pretty much any context you might want to fetch resources in.

> The fetch() method takes one mandatory argument, the path to the resource you want to fetch. It returns a Promise that resolves to the Response to that request, whether it is successful or not.

So we can use it like:

```js
fetch(...).then(...).catch(...);
```

Or by using `async/await`.

https://dev.to/johnpaulada/synchronous-fetch-with-asyncawait

damienomurchu: Resolving. Hadn't been sure that `.text()` returns a promise, but [it does](https://developer.mozilla.org/en-US/docs/Web/API/Body/text) so await required.

ui/src/DataService.js
```diff
@@ -8,7 +8,7 @@ import config from './config/config';
  * @param {object|Array} body - JSON Payload to send in the request
  * @returns a config object to send in HTTP requests
  */
-const requestConfig = (method, body) => ({
+export const requestConfig = (method, body) => ({
```
damienomurchu: Can this be exported [via the `dataService` object](https://github.com/aerogear/mobile-security-service/blob/2062561c1fdab518763fa2dc91d7c07cdcc2f85b/ui/src/DataService.js#L90-L94) so we are consistent

ui/src/containers/App.js
damienomurchu: Re-opening this for further clarification - if our approach here to disable the back button so the user cannot navigate back to that page, what stops the user from manually entering the URL & navigating back to that view and being able to see the UI?

- it seems disabling the back button is security by obscurity which is not recommended

ui/src/DataService.js
craicoverflow: It could, but the reason I'm exporting it separately is because the AuthService does not want the entire DataService. If I export it here then the AuthService would have to import the DataService just to use the `requestConfig`.

PR #155, opened 2019-04-18
Show an alert when the user saves an app or when the save failed

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9045

## What

Added a Patternfly 4 Alert when an app is successfully saved or when the save failed. 

I wrapped the PF Alert in a custom component, as the PF Alert does not have any props available to control the visibility of the Alert.

## Why

To give the user visual feedback when the save has finished.

## Verification Steps
 
1. Go to the App Page.
2. Make changes to one of the versions.
3. Click `Save`.
4. You should see an Alert in the top right that reads "Save successful".
5. Click the `x` on the Alert. The Alert should now disappear.
6. Temporarily modify the [`UpdateAppVersions`](https://github.com/aerogear/mobile-security-service/blob/master/pkg/web/apps/apps_http_handler.go#L113) handler in the server to return a Bad Request.
7. Repeat steps **1-3**.
8. You should see an Alert in the top right that reads "Save failed".

## Checklist:

## Progress

- [ ] Finished task
- [ ] TODO

### DIFF
--- a/ui/src/components/common/CustomAlert/CustomAlert.js
+++ b/ui/src/components/common/CustomAlert/CustomAlert.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import './CustomAlert.css';
+import PropTypes from 'prop-types';
+import { Alert, AlertActionCloseButton } from '@patternfly/react-core';
+
+/**
+ * A custom alert component that wraps the Patternfly 4 Alert.
+ *
+ * @param {Object} visible - The visible state of the alert.
+ * @param {func} onClose - The callback function to execute on close.
+ * @param {*} children - Any child props - usually the description.
+ * @param {Object} props - The component props
+ */
+const CustomAlert = ({ visible, onClose, children, ...props }) => {
+  if (!visible) {
+    return null;
+  }
+
+  return (
+    <Alert
+      className='Alert'
+      {...props}
+      action={<AlertActionCloseButton onClose={onClose} />}>
+      {children}
+    </Alert>
+  );
+};
+
+CustomAlert.propTypes = {
+  visible: PropTypes.bool.isRequired,
+  onClose: PropTypes.func.isRequired,
+  children: PropTypes.any
+};
+
+export default CustomAlert;
--- a/ui/src/containers/appView/AppPageContainer.js
+++ b/ui/src/containers/appView/AppPageContainer.js
@@ -20,6 +20,7 @@ import {
   saveAppVersions
 } from '../../actions/actions-ui';
 import AppService from '../../services/appService';
+import CustomAlert from '../../components/common/CustomAlert/CustomAlert';
 
 /**
  * Stateful container component for the AppPage
@@ -28,6 +29,10 @@ import AppService from '../../services/appService';
  * @extends {React.Component}
  */
 export class AppPageContainer extends React.Component {
+  state = {
+    alert: { visible: false }
+  }
+
   componentWillMount () {
     this.props.getAppById(this.props.match.params.id);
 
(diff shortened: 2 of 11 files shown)


craicoverflow:
I have left the error message generic as I don't foresee any errors that can be rectified by the user (missing or invalid data). Thoughts?

ui/src/components/common/CustomAlert/CustomAlert.js
```diff
@@ -0,0 +1,41 @@
...
+      {...props}
+      action={<AlertActionCloseButton onClose={onClose} />}>
+      {children}
+    </Alert>
+  );
+};
+
+CustomAlert.propTypes = {
+  visible: PropTypes.bool.isRequired,
+  onClose: PropTypes.func,
```
davidffrench: Do you think this should be a required proptype?

ui/src/components/common/CustomAlert/CustomAlert.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param {Object} props - The component props
+ */
+const CustomAlert = ({ visible, onClose, children, ...props }) => {
+  const [isVisible, setVisibility] = useState(visible);
+
+  useEffect(() => {
+    setVisibility(visible);
+  }, [visible]);
+
+  if (!isVisible) {
```
davidffrench:
```suggestion
  if (! visible) {
```
Lines 15 - 20 do not look to be needed. You can switch the conditional to just use the suggestion above.

ui/src/containers/appView/AppPageContainer.js
```diff
@@ -42,6 +47,45 @@ export class AppPageContainer extends React.Component {
...
+
+  /**
+   * Show an Alert
+   *
+   * @param {string} title - The alert title
+   * @param {*} variant - one of danger, success, warning, info
+   * @param {*} [description] - displays an optional Alert body
+   * @param {number} [timeout] - optionally set a duration of time before the alert disappears
+   */
+  showAlert (title, variant, description = null, timeout = 5000) {
```
davidffrench: `timeout` is no longer required here.

ui/src/containers/appView/AppPageContainer.js
```diff
@@ -42,6 +47,45 @@ export class AppPageContainer extends React.Component {
...
+    }
+  }
+
+  /**
+   * Show an Alert
+   *
+   * @param {string} title - The alert title
+   * @param {*} variant - one of danger, success, warning, info
+   * @param {*} [description] - displays an optional Alert body
+   * @param {number} [timeout] - optionally set a duration of time before the alert disappears
```
davidffrench: `timeout` is no longer required here.

ui/src/containers/appView/AppPageContainer.js
craicoverflow: My bad, never pushed changes with it gone.

ui/src/containers/appView/AppPageContainer.js
craicoverflow: My bad, never pushed changes with it gone.

ui/src/components/common/CustomAlert/CustomAlert.js
craicoverflow: Yes it should be.

ui/src/components/common/CustomAlert/CustomAlert.js
craicoverflow: Yeah, I don't know why I thought I needed this :joy:

davidffrench: was probably needed with the timeout logic.

PR #156, opened 2019-06-04
feat(CI): update CI to push images to quay.io instead of dockerhub

damienomurchu:
## Motivation
Host images on quay.io instead of dockerhub

## What
Host images on quay.io instead of dockerhub

## Why
Aerogear images to be hosted on quay.io

## How
Update push command in makefile & add quay.io credentials to project in CircleCI dashboard

## Verification Steps
- deploy to your local environment
```
minishift start
oc login -u system:admin
make create-all
make create-app
``` 
- update MSS image to point to testing image: `quay.io/damienomurchu/mobile-security-service:c29f31e`
- check that pod redeploys correctly
- check that no functionality has been broken (JIRA [here](https://issues.jboss.org/browse/AEROGEAR-9233) is a good reference as to what to check)
- check that the UI renders the app(s) in the table and the logged in user in the header correctly

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin-release:golang-1.10 as builder
+FROM openshift/origin-release:golang-1.12 as builder
 WORKDIR /go/src/github.com/aerogear/mobile-security-service
 COPY . .
 RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh \
--- a/Makefile
+++ b/Makefile
@@ -12,10 +12,11 @@ BINARY ?= mobile-security-service
 # This follows the output format for goreleaser
 BINARY_LINUX_64 = ./dist/linux_amd64/$(BINARY)
 
-DOCKER_LATEST_TAG = $(ORG_NAME)/$(APP_NAME):latest
-DOCKER_MASTER_TAG = $(ORG_NAME)/$(APP_NAME):master
+IMAGE_REGISTRY=quay.io
+DOCKER_LATEST_TAG = $(IMAGE_REGISTRY)/$(ORG_NAME)/$(APP_NAME):latest
+DOCKER_MASTER_TAG = $(IMAGE_REGISTRY)/$(ORG_NAME)/$(APP_NAME):master
 RELEASE_TAG ?= $(CIRLE_TAG)
-DOCKER_RELEASE_TAG = $(ORG_NAME)/$(APP_NAME):$(RELEASE_TAG)
+DOCKER_RELEASE_TAG = $(IMAGE_REGISTRY)/$(ORG_NAME)/$(APP_NAME):$(RELEASE_TAG)
 
 LDFLAGS=-ldflags "-w -s -X main.Version=${TAG}"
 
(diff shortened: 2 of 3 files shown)


damienomurchu:
Thanks for the quick approval @camilamacedo86 - PR to update operator is here https://github.com/aerogear/mobile-security-service-operator/pull/113 if you can review too

damienomurchu:
Just running some small checks before I merge, fyi

Makefile
```diff
@@ -65,13 +66,13 @@ docker_build_master:
 
 .PHONY: docker_push_release
 docker_push_release:
-	@docker login --username $(DOCKERHUB_USERNAME) --password $(DOCKERHUB_PASSWORD)
+	@docker login --username $(QUAY_USERNAME) --password $(QUAY_PASSWORD) quay.io
```
laurafitzgerald: We could use IMAGE_REGISTRY variable here instea dof hardcoding `quay.io`

Makefile
```diff
@@ -65,13 +66,13 @@ docker_build_master:
...
 	docker push $(DOCKER_LATEST_TAG)
 	docker push $(DOCKER_RELEASE_TAG)
 
 .PHONY: docker_push_master
 docker_push_master:
-	@docker login --username $(DOCKERHUB_USERNAME) --password $(DOCKERHUB_PASSWORD)
+	@docker login --username $(QUAY_USERNAME) --password $(QUAY_PASSWORD) quay.io
```
laurafitzgerald: same here

Makefile
```diff
@@ -65,13 +66,13 @@ docker_build_master:
 
 .PHONY: docker_push_release
 docker_push_release:
-	@docker login --username $(DOCKERHUB_USERNAME) --password $(DOCKERHUB_PASSWORD)
+	@docker login --username $(QUAY_USERNAME) --password $(QUAY_PASSWORD) quay.io
 	docker push $(DOCKER_LATEST_TAG)
```
laurafitzgerald: Why do we push latest and release tag both here for `docker_push_release`?

Makefile
```diff
@@ -65,13 +66,13 @@ docker_build_master:
 
 .PHONY: docker_push_release
```
laurafitzgerald: can we change the name of these commands?.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openshift/origin-release:golang-1.10 as builder
+FROM openshift/origin-release:golang-1.12 as builder
```
laurafitzgerald: What level of testing do we need to do in order to ensure a change to the go version is okay?

Makefile
damienomurchu: This is to publish an image with the `<release-tag>`, and also to publish an image with the `latest` tag, ie so you can either pull a specific version, or just the latest tagged version

Makefile
damienomurchu: Its possible. I would not see `docker_push_release` referring to the registry but rather the way in which we push the image
/2cents

Dockerfile
damienomurchu: Good question. For this PR I will verify all functionality of the MSS after updating to this image so as to ensure no broken functionality.

damienomurchu:
@laurafitzgerald think I've addressed all of your requested changes and have also checked that no functionality has been broken by the update to use the images from `quay.io` or the update to use a later base image for the mss image

If you / @camilamacedo86 could re-review that would be awesome :)

damienomurchu:
Thanks @davidffrench @camilamacedo86 @laurafitzgerald - just need an approval now if you don't mind doing the honours (to confirm: I won't be merging this until later, so you can feel free to approve now)

PR #165, opened 2019-08-08
fix: ensure app ID check is case insensitive.

craicoverflow:
https://issues.jboss.org/browse/AEROGEAR-9722

This change ensures that the app ID unique constraint is case sensitive.

## Verification

1. Create an app:

```json
POST http://localhost:3001/api/apps
{
	"deviceId": "a2895cc1-28d7-4283-932d-8bcab9e1b566",
	"deviceVersion": "3.7",
	"version": "2",
	"deviceType": "Android",
	"appId": "com.aerogear.testapp"
}
```

2. Create another app with same ID but with different cases:

```json
POST http://localhost:3001/api/apps
{
	"deviceId": "a2895cc1-28d7-4283-932d-8bcab9e1b566",
	"deviceVersion": "3.7",
	"version": "2",
	"deviceType": "Android",
	"appId": "com.aerogear.testapP"
}
```

3. You should get a 409 Conflict error.

```json
{
    "message": "Your Item already exists",
    "statusCode": 409
}
```

### DIFF
--- a/pkg/web/apps/apps_psql_repository.go
+++ b/pkg/web/apps/apps_psql_repository.go
@@ -2,6 +2,7 @@ package apps
 
 import (
 	"database/sql"
+	"strings"
 
 	"time"
 
@@ -192,10 +193,10 @@ func (a *appsPostgreSQLRepository) GetDeviceByVersionAndAppID(version string, ap
 func (a *appsPostgreSQLRepository) GetAppByAppID(appID string) (*models.App, error) {
 	app := models.App{}
 
-	sqlStatement := `SELECT id,app_id,app_name,deleted_at FROM app WHERE app_id=$1;`
+	sqlStatement := `SELECT id,app_id,app_name,deleted_at FROM app WHERE LOWER(app_id)=$1;`
 
 	var deletedAt sql.NullString
-	err := a.db.QueryRow(sqlStatement, appID).Scan(&app.ID, &app.AppID, &app.AppName, &deletedAt)
+	err := a.db.QueryRow(sqlStatement, strings.ToLower(appID)).Scan(&app.ID, &app.AppID, &app.AppName, &deletedAt)
 	app.DeletedAt = deletedAt.String
 
 	if err != nil {
@@ -213,9 +214,9 @@ func (a *appsPostgreSQLRepository) GetAppByAppID(appID string) (*models.App, err
 func (a *appsPostgreSQLRepository) GetActiveAppByAppID(appID string) (*models.App, error) {
 	app := models.App{}
 
-	sqlStatement := `SELECT id,app_id,app_name FROM app WHERE app_id=$1 AND deleted_at IS NULL;`
+	sqlStatement := `SELECT id,app_id,app_name FROM app WHERE LOWER(app_id)=$1 AND deleted_at IS NULL;`
 
-	err := a.db.QueryRow(sqlStatement, appID).Scan(&app.ID, &app.AppID, &app.AppName)
+	err := a.db.QueryRow(sqlStatement, strings.ToLower(appID)).Scan(&app.ID, &app.AppID, &app.AppName)
 
 	if err != nil {
 		log.Error(err)
@@ -295,7 +296,7 @@ func (a *appsPostgreSQLRepository) DisableAllAppVersionsAndSetDisabledMessageByA
 	_, err := a.db.Exec(`
 		UPDATE version
 		SET disabled_message=$1,disabled=True
-		WHERE app_id=$2;`, message, appID)
+		WHERE LOWER(app_id)=$2;`, message, strings.ToLower(appID))
 
 	if err != nil {
 		log.Error(err)
@@ -312,7 +313,7 @@ func (a *appsPostgreSQLRepository) DisableAllAppVersionsByAppID(appID string) er
 	_, err := a.db.Exec(`
 		UPDATE version
 		SET disabled=True
-		WHERE app_id=$1;`, appID)
+		WHERE LOWER(app_id)=$1;`, strings.ToLower(appID))
 
 	if err != nil {
 		log.Error(err)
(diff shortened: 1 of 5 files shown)


laurafitzgerald:
👀

laurafitzgerald:
Verification Steps Taken:
`make setup`
`docker-compose up -db`
`make serve`
Then steps as per the description

@craicoverflow I'm getting some unusual errors here. Could we chat about it when you have some time?

craicoverflow:
> Verification Steps Taken:
> `make setup`
> `docker-compose up -db`
> `make serve`
> Then steps as per the description
> 
> @craicoverflow I'm getting some unusual errors here. Could we chat about it when you have some time?

Of course.

pkg/web/apps/apps_psql_repository.go
```diff
@@ -193,7 +193,7 @@ func (a *appsPostgreSQLRepository) GetDeviceByVersionAndAppID(version string, ap
 func (a *appsPostgreSQLRepository) GetAppByAppID(appID string) (*models.App, error) {
 	app := models.App{}
 
-	sqlStatement := `SELECT id,LOWER(app_id),app_name,deleted_at FROM app WHERE LOWER(app_id)=$1;`
+	sqlStatement := `SELECT id,app_id,app_name,deleted_at FROM app WHERE LOWER(app_id)=$1;`
```
camilamacedo86: I understand that is missing this place: https://github.com/aerogear/mobile-security-service/blob/57e14c8d31ee55843423f09615e1b88bc3972927/pkg/web/apps/apps_psql_repository.go#L112

craicoverflow: Thank you :smile:

craicoverflow: After investigating further, the line you quoted is getting an app by its ID (PK, uuid), not the app ID (string).

craicoverflow: I think you intended to reference line 227:

https://github.com/aerogear/mobile-security-service/blob/57e14c8d31ee55843423f09615e1b88bc3972927/pkg/web/apps/apps_psql_repository.go#L217

camilamacedo86: Hi @craicoverflow has the GetActiveAppID which is by the appID. Could you please check this one? It is that one which I am saying that is missing this change.

craicoverflow: Yep I am updating this now.

Boomatang:
This does solve the issue with case insensitive. But a second issue now is once you delete the entry `DELETE http://localhost:3001/api/apps/{id}` you can not create a new binding using the old id.

craicoverflow:
> This does solve the issue with case insensitive. But a second issue now is once you delete the entry `DELETE http://localhost:3001/api/apps/{id}` you can not create a new binding using the old id.

I feel like this is a separate issue that probably existed already. Can you check it on master?

camilamacedo86:
Hi @Boomatang, 

IMHO: We do not need care over the old one since it was not released yet. So, nobody will have the old one. Which we need to ensure is that the create/delete/update app and init will work as expected after this change. 

@craicoverflow

Boomatang:
> > This does solve the issue with case insensitive. But a second issue now is once you delete the entry `DELETE http://localhost:3001/api/apps/{id}` you can not create a new binding using the old id.
> 
> I feel like this is a separate issue that probably existed already. Can you check it on master?

I have check this issue on master. On the current master branch when an entry is added with an existing soft deleted app Id, the delete_at field gets set to null. There is not a new entry, the old entry is reused.

craicoverflow:
> Following what need be changed here IMHO
> 
>     * all places where insert/select/delete/update by app_id need to be changed to use the lower case

I don't think this should be changed. A developer may prefer their app ID to have uppercase letters. What is the reason for persisting in lowercase?

craicoverflow:
> > > This does solve the issue with case insensitive. But a second issue now is once you delete the entry `DELETE http://localhost:3001/api/apps/{id}` you can not create a new binding using the old id.
> > 
> > 
> > I feel like this is a separate issue that probably existed already. Can you check it on master?
> 
> I have check this issue on master. On the current master branch when an entry is added with an existing soft deleted app Id, the delete_at field gets set to null. There is not a new entry, the old entry is reused.

Yes originally the CreateApp function would "undelete" a soft-deleted app, but I changed this without realising. This has been reverted now.

camilamacedo86:
> I don't think this should be changed. A developer may prefer their app ID to have uppercase letters. What is the reason for persisting in lowercase?

So, it will allow having in the DB one with upper case and other with not but the selects always will return the lower one. Is it something that really makes sense?

craicoverflow:
> > I don't think this should be changed. A developer may prefer their app ID to have uppercase letters. What is the reason for persisting in lowercase?
> 
> So, it will allow having in the DB one with upper case and other with not but the selects always will return the lower one. Is it something that really makes sense?

We will not have one ID with uppercase and one app ID with lowercase as this will be checked and prevented, they will be treated as the same app.

https://github.com/aerogear/mobile-security-service/blob/edf2c0884ec0731bb99d7780b759c60cdb034dba/pkg/web/apps/apps_service.go#L135

pkg/web/apps/apps_service_test.go
```diff
@@ -404,7 +404,7 @@ func Test_appsService_UpdateAppVersions(t *testing.T) {
 	}
 }
 
-func Test_appsService_CreateUpdateApp(t *testing.T) {
+func Test_appsService_CreateApp(t *testing.T) {
```
camilamacedo86: Hi @craicoverflow you change the test to check just to create a new app.  However, we need to test the update as well. How be sure that this change will not affect it?

pkg/web/apps/apps_service_test.go
```diff
@@ -534,7 +534,7 @@ func Test_appsService_CreateUpdateApp(t *testing.T) {
 			err := a.CreateApp(*tt.data)
 
 			if (err != nil) && (tt.wantErr != err || tt.wantErr == nil) {
-				t.Errorf("appsService.BindingAppByApp() error = %v, wantErr %v", err, tt.wantErr)
+				t.Errorf("appsService.CreateApp() error = %v, wantErr %v", err, tt.wantErr)
```
camilamacedo86: why we the BindingAppByApp was change to use the CreateApp?

PR #168, opened 2019-08-13
docs: remove extra line from readme

craicoverflow:


.circleci/config.yml
```diff
@@ -28,8 +28,7 @@ jobs:
       - run: make build
       - run: make test-unit
       - run: make test-integration-cover
-      # fixme: removed this while coveralls is down. Re-add in future
```
camilamacedo86: It is a comment for we know why the line is commented. We need to come back it gain when the coveralls be fixed.

PR #169, opened 2019-08-13
fix(ci) : upgrade node and golang versions

camilamacedo86:
## Motivation
- Issue faced when the CI is building UI (integrity checksum failed when using sha1:)

## What
- Fix `integrity checksum failed when using sha1:` issue
- Upgrade node and go versions in the CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
     working_directory: /go/src/github.com/aerogear/mobile-security-service
 
     docker:
-      - image: circleci/golang:1.10
+      - image: circleci/golang:1.12
       - image: registry.access.redhat.com/rhscl/postgresql-96-rhel7:latest
         ports:
           - 5432:5432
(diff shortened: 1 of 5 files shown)


ui/package.json
```diff
@@ -9,7 +9,7 @@
     "@patternfly/react-core": "2.3.6",
     "@patternfly/react-table": "1.0.7",
     "dotenv": "^6.2.0",
-    "lodash": "4.17.11",
+    "lodash": "^4.17.15",
```
craicoverflow:
```suggestion
"lodash": "4.17.15",
```

ui/package.json
```diff
@@ -9,7 +9,7 @@
...
     "lodash-es": "4.17.11",
```
craicoverflow: Should this also be increased to 4.7.15 also?

.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     working_directory: ~/mobile-security-service
 
     docker:
-      - image: circleci/node:8.15.0
+      - image: circleci/node:12
```
craicoverflow: I believe we should set the Node version to 10. See that the Dockerfile also uses Node v10:

https://github.com/aerogear/mobile-security-service/blob/c29f31e90d667da1f01f1664aa15bf1dc5ef67f8/Dockerfile#L8

ui/package.json
camilamacedo86: It will not work and will cause the issue.

ui/package.json
camilamacedo86: I do not think so.

.circleci/config.yml
camilamacedo86: 👍 doing

